Issue #4, opened 2016-12-02
Clarify manual submission doc in Usage.md


lisaong:
Minor clarification to the wordings in "Manual Submissions" around patching in values from the json file into the submission payload.

lisaong:
Working on a proposed change

PR #5, opened 2016-12-02
Issue #4 - Clarify manual submission doc in Usage.md

lisaong:
#4

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -225,7 +225,7 @@ function Format-SimpleTableString
     }
 }
 
-function DeepCopy-Object
+function Copy-ObjectDeep
 <#
     .SYNOPSIS
         Creates a deep copy of a serializable object.
@@ -253,7 +253,7 @@ function DeepCopy-Object
 #>
 {
     [CmdletBinding()]
-    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseApprovedVerbs", "", Justification="Intentional.  This isn't exported, and needed to be explicit relative to Copy-Object.")]
+    [Alias('DeepCopy-Object')]
     param(
         [Parameter(Mandatory)]
         [PSCustomObject] $Object
--- a/Documentation/USAGE.md
+++ b/Documentation/USAGE.md
@@ -260,28 +260,37 @@ manually.
 
  * Clone the existing published submission so that you can generate an update.
  
-       $sub = New-ApplicationSubmission -IapId <iapId> [-Force]
+          $sub = New-ApplicationSubmission -AppId <appId> [-Force]
 
     * By using the `-Force` switch, it will call `Remove-ApplicationSubmission` behind the
       scenes if it finds that there's an existing pending submission for your app.
 
  * Read in the content of the json file from your `New-SubmissionPackage` payload:
  
-       $json = (Get-Content .\submission.json -Encoding UTF8) | ConvertFrom-Json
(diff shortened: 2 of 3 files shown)


lisaong:
Here's the rendered view: https://github.com/lisaong/StoreBroker/blob/manual_submission_docs/Documentation/USAGE.md

lisaong:
See: https://github.com/lisaong/StoreBroker/blob/manual_submission_docs/Documentation/USAGE.md

HowardWolosky-MSFT:
Thanks for doing this, Lisa!

Overall, this looks good.  The main issue with this is that `DeepCopy-Command` is not currently an exported function, so the second example will never work for anyone.  We could consider exporting it (via `StoreBroker\StoreBroker.psd1`), but we'd also have to make a change to its declaration in `Helpers.ps1` to make its default name `Copy-ObjectDeep` with an _alias_ of `DeepCopy-Object` to avoid any module import warnings of using unsupported nouns.

One other request, while you're in that section, is to fix the code in the first bulletpoint.  Not sure why I had `-IapId` in there.  That should probably be `-AppId <appId>`.

lisaong:
Done and done.  I tested the export of DeepCopy-Object by comparing the behavior before and after the export.

HowardWolosky-MSFT:
Great, thanks for the update.

Since this is a documentation update for clarity, I'm going to be a bit more picky here....I'm noticing one other thing in the documentation now too, and it's not your fault, but hopefully you can fix it.  The example code in the bullet points is not getting formatted as a code snippet (not in your new bullet points, nor in the other existing bullet points within that section).  I think yours need another 3 spaces indentation before they get recognized as code snippets.  In other markdown editors, I'm seeing the existing bullet point code snippets show up right, but not when displayed in GitHub for some reason.  Thoughts?

Also, as I'm not sure if I can modify your pull request when I accept it, can you go ahead and update the version number in `StoreBroker.psd1` to be `1.0.1` to reflect the minor change made to `Helpers.ps1` and the `psd` file?

Thanks once more.

StoreBroker/Helpers.ps1
```diff
@@ -253,6 +253,7 @@ function DeepCopy-Object
 #>
 {
     [CmdletBinding()]
+    [Alias('DeepCopy-Object')]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseApprovedVerbs", "", Justification="Intentional.  This isn't exported, and needed to be explicit relative to Copy-Object.")]
```
HowardWolosky-MSFT: Now that you're using an approved verb by default, we no longer need to suppress this code analysis message since it won't show up.  Please remove this, and then run [static analysis](https://github.com/Microsoft/StoreBroker/blob/master/CONTRIBUTING.md#static-analysis) again and make sure it's still clean.

lisaong:
Updated and re-squashed commits.  There were a bunch of existing warnings from Invoke-ScriptAnalyzer that were unrelated to my changes, I diff'ed of the output of the original pre-renamed version without the suppression and verified that that specific warning around CopyObject went away.

HowardWolosky-MSFT:
This final update looks great.  I'll approve it and pull it in.  I'm interested in hearing more about the static analysis warnings that you were seeing.  It was clean before I submitted v1.0.0 to GitHub, and I just ran it again on what's in master and it's still clean.  I also ran it against your branch and it was clean.  It's possible that you had other modules loaded at the time that you ran the static analysis on StoreBroker which resulted in you getting some false positives.

PR #13, opened 2017-01-25
skip console messages if the host is non-interactive

### LINKED ISSUE #12: Console messages should be suppressed in Write-Log for non-interactive consoles
Write-Log currently outputs both to the console and the text log file.  

For some server applications, store broker can be invoked through a non-interactive console, which causes Write-Log to fail if console logging is enabled for a given log level (typically the default or "Info" level and above). 

If the console is non-interactive, there is no point logging to it anyway, so console output should be suppressed.

lisaong:
Issue #12 

First try the -NonInteractive parameter check (which should catch most cases).  If no parameters are supplied, check for the 'Default Host' name.  

The latter check is needed when running the scripts within the [EdgePS](https://github.com/tjanczuk/edge#how-to-script-powershell-in-a-nodejs-application) hosting environment. Unfortunately, no parameters are supplied in the binary version that we are consuming.

Resolves #12: Console messages should be suppressed in Write-Log for non-interactive consoles

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -105,7 +105,6 @@ function Wait-JobWithAnimation
         set of configuration options that Wait-Job does.
 #>
     [CmdletBinding()]
-    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This function is intended for human interaction, not for scripting.  Write-Host makes the most sense for visible feedback.")]
     Param(
         [Parameter(Mandatory)]
         [string] $JobName,
@@ -125,27 +124,27 @@ function Wait-JobWithAnimation
     $iteration = 0
     while (((Get-Job -Name $JobName).state -eq 'Running'))
     {
-        Write-Host "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
+        Write-InteractiveHost "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
         Start-Sleep -Milliseconds ([int](1000/$framesPerSecond))
         $iteration++
     }
 
     if ((Get-Job -Name $JobName).state -eq 'Completed')
     {
-        Write-Host "`rDONE - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Green
+        Write-InteractiveHost "`rDONE - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Green
 
         # We forcibly set Verbose to false here since we don't need it printed to the screen, since we just did above -- we just need to log it.
         Write-Log "DONE - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -Level Verbose -Verbose:$false
     }
     else
     {
-        Write-Host "`rDONE (FAILED) - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Red
+        Write-InteractiveHost "`rDONE (FAILED) - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Red
 
         # We forcibly set Verbose to false here since we don't need it printed to the screen, since we just did above -- we just need to log it.
         Write-Log "DONE (FAILED) - Operation took $([int]($iteration / $framesPerSecond)) second(s) $Description" -Level Verbose -Verbose:$false
     }
 
-    Write-Host ""
+    Write-InteractiveHost ""
 }
 
 function Format-SimpleTableString
@@ -401,7 +400,6 @@ function Write-Log
     [CmdletBinding(SupportsShouldProcess)]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidGlobalVars", "", Justification="We use global variables sparingly and intentionally for module configuration, and employ a consistent naming convention.")]
-    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="We'd like to use Write-Information instead, but it's not supported on PS 4.0 which we need to support.")]
     param(
         [Parameter(
             Mandatory,
@@ -442,21 +440,21 @@ function Write-Log
 
         switch ($Level)
         {
-            'Error'   { Write-Error $ConsoleMessage }
-            'Warning' { Write-Warning $ConsoleMessage }
-            'Verbose' { Write-Verbose $ConsoleMessage }
-            'Debug'   { Write-Debug $ConsoleMessage }
+            'Error'   { Write-Error $consoleMessage }
+            'Warning' { Write-Warning $consoleMessage }
+            'Verbose' { Write-Verbose $consoleMessage }
+            'Debug'   { Write-Debug $consoleMessage }
             'Info'    {
                 # We'd prefer to use Write-Information to enable users to redirect that pipe if
                 # they want, unfortunately it's only available on v5 and above.  We'll fallback to
                 # using Write-Host for earlier versions (since we still need to support v4).
                 if ($PSVersionTable.PSVersion.Major -ge 5)
                 {
-                    Write-Information $ConsoleMessage -InformationAction Continue
+                    Write-Information $consoleMessage -InformationAction Continue
                 }
                 else
                 {
-                    Write-Host $ConsoleMessage
+                    Write-InteractiveHost $consoleMessage
                 }
             }
         }
@@ -479,7 +477,7 @@ function Write-Log
                 # Let's do best effort here and if we can't log something, just report
                 # it and move on.
                 $output += "This is non-fatal, and your command will continue.  Your log file will be missing this entry:"
-                $output += $ConsoleMessage
+                $output += $consoleMessage
                 Write-Warning ($output -join [Environment]::NewLine)
             }
             else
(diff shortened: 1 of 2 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -71,6 +71,14 @@ function Initialize-HelpersGlobalVariables
     {
         $global:SBNotifyCredential = [PSCredential]$null
     }
+
+    # Initialize and cache whether this console is interactive
+    $global:SBInInteractiveConsole = ![Bool]([Environment]::GetCommandLineArgs() -like '-noni*')
+    if ($global:SBInInteractiveConsole)
+    {
+        # Some hosts like "Default Host" don't take command line arguments, but default to non-interactive
+        $global:SBInInteractiveConsole = (Get-Host).Name -ne 'Default Host'
```
HowardWolosky-MSFT: I'm curious if you can avoid this check if you do a check for `[Environment]::UserInteractive` and `-and` that with your other check.  I was looking at [StackOverflow](http://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode) and that appeared to be the consensus for this issue.

StoreBroker/Helpers.ps1
```diff
@@ -120,27 +128,37 @@ function Wait-JobWithAnimation
     $iteration = 0
     while (((Get-Job -Name $JobName).state -eq 'Running'))
     {
-        Write-Host "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
+        if ($global:SBInInteractiveConsole)
+        {
+            Write-Host "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
+        }
```
danbelcher-MSFT: Seems like we could have a Write-InteractiveHost function that takes care of checking if we are in an interactive host and noops if we aren't.  Then you just need to change the function names here.

StoreBroker/Helpers.ps1
```diff
@@ -383,23 +401,27 @@ function Write-Log
             $env:username,
             $Message
 
-        switch ($Level)
+        # Output to the console if it is interactive
+        if ($global:SBInInteractiveConsole)
```
danbelcher-MSFT: Could solve this with Write-InteractiveHost, mentioned above

StoreBroker/Helpers.ps1
HowardWolosky-MSFT: That's a great suggestion.  +1

StoreBroker/Helpers.ps1
lisaong: Tried that, and it unfortunately didn't work (returned true in the environment we are using), so I kept the 'Default Host' check.

StoreBroker/Helpers.ps1
lisaong: Fixed

StoreBroker/Helpers.ps1
```diff
@@ -418,12 +440,15 @@ function Write-Log
 
             if (Test-Path -Path $Path -PathType Leaf)
             {
-                # The file exists, but likely is being held open by another process.
-                # Let's do best effort here and if we can't log something, just report
-                # it and move on.
-                $output += "This is non-fatal, and your command will continue.  Your log file will be missing this entry:"
-                $output += $ConsoleMessage
-                Write-Warning ($output -join [Environment]::NewLine)
+                if ($global:SBInInteractiveConsole)
```
lisaong: Reverted

StoreBroker/Helpers.ps1
lisaong: Done.  I ended up generating a proxy function, ripping out the guts (because it's more complicated than our purposes), and then forwarding the call to Write-Host appropriately.

In case you are curious, the proxy function is this complex because it allows you to augment parameters.  I pasted the original below.

https://blogs.technet.microsoft.com/heyscriptingguy/2011/03/01/proxy-functions-spice-up-your-powershell-core-cmdlets/


    [CmdletBinding(HelpUri='http://go.microsoft.com/fwlink/?LinkID=113426', RemotingCapability='None')]
    param(
        [Parameter(Position=0, ValueFromPipeline=$true, ValueFromRemainingArguments=$true)]
        [System.Object]
        ${Object},

        [switch]
        ${NoNewline},

        [System.Object]
        ${Separator},

        [System.ConsoleColor]
        ${ForegroundColor},

        [System.ConsoleColor]
        ${BackgroundColor})

    begin
    {
        try {
            $outBuffer = $null
            if ($PSBoundParameters.TryGetValue('OutBuffer', [ref]$outBuffer))
            {
                $PSBoundParameters['OutBuffer'] = 1
            }
            $wrappedCmd = $ExecutionContext.InvokeCommand.GetCommand('Microsoft.PowerShell.Utility\Write-Host', [System.Management.Automation.CommandTypes]::Cmdlet)
            $scriptCmd = {& $wrappedCmd @PSBoundParameters }
            $steppablePipeline = $scriptCmd.GetSteppablePipeline($myInvocation.CommandOrigin)
            $steppablePipeline.Begin($PSCmdlet)
        } catch {
            throw
        }
    }

    process
    {
        try {
            $steppablePipeline.Process($_)
        } catch {
            throw
        }
    }

    end
    {
        try {
            $steppablePipeline.End()
        } catch {
            throw
        }
    }
    <#

    .ForwardHelpTargetName Microsoft.PowerShell.Utility\Write-Host
    .ForwardHelpCategory Cmdlet

    #>
}

StoreBroker/Helpers.ps1
```diff
@@ -105,7 +105,6 @@ function Wait-JobWithAnimation
         set of configuration options that Wait-Job does.
 #>
     [CmdletBinding()]
-    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This function is intended for human interaction, not for scripting.  Write-Host makes the most sense for visible feedback.")]
```
lisaong: re-ran  Invoke-ScriptAnalyzer -Path .\ -Recurse, and no new hits for Helpers.ps1 due to my change.  @HowardWolosky-MSFT, I do see this new warning: 

PSUseDeclaredVarsMoreThanAssignment Warning      Helpers.ps 77    The variable 'SBUseUTC' is assigned but
s                                                1                never used.

StoreBroker/Helpers.ps1
lisaong: Unfortunately, [Environment]::UserInteractive returned true in the edge-ps environment, so I had to keep the check.

StoreBroker/Helpers.ps1
lisaong: Done. Thanks for the feedback!

StoreBroker/Helpers.ps1
lisaong: Removed. Thanks!

StoreBroker/Helpers.ps1
lisaong: +1, done.

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,58 @@ function Send-SBMailMessage
...
+        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+
+    .EXAMPLE
+        Write-InteractiveHost "Test"
+        Write-InteractiveHost "Test" -NoNewline -f Yellow
+#>
+
+    [CmdletBinding(HelpUri='http://go.microsoft.com/fwlink/?LinkID=113426', RemotingCapability='None')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This provides a wrapper around Write-Host. In general, we'd like to use Write-Information, but it's not supported on PS 4.0 which we need to support.")]
+    param(
```
danbelcher-MSFT: nit picky stuff, this should be reformatted to be consistent with the file
- closing paren ')' for the param block should be on its own line and aligned under the 'p' of 'param'
- variables should be on the same line as their type, separated by a space
- variables don't need the curlys, just the $ and their name
- each characteristic in the Parameter attribute should be on its own line, see ConvertTo-Array in this file
- just specify the characteristic, no need to add the $true as it is implicit.  For example, it's enough to say 'ValueFromPipeline'.  See ConvertTo-Array in this file

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,58 @@ function Send-SBMailMessage
...
+    # Determine if the host is interactive
+    if ([Environment]::UserInteractive -and ![Bool]([Environment]::GetCommandLineArgs() -like '-noni*') -and (Get-Host).Name -ne 'Default Host')
+    {
+        # Special handling for OutBuffer (generated for the proxy function)
+        $outBuffer = $null
+        if ($PSBoundParameters.TryGetValue('OutBuffer', [ref]$outBuffer))
+        {
+            $PSBoundParameters['OutBuffer'] = 1
+        }
+        Write-Host @PSBoundParameters
```
danbelcher-MSFT: nit: add a newline after the if-block before the next statement.

StoreBroker/Helpers.ps1
lisaong: Done!

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,54 @@ function Send-SBMailMessage
...
+
+        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+
+    .EXAMPLE
+        Write-InteractiveHost "Test"
+        Write-InteractiveHost "Test" -NoNewline -f Yellow
+#>
+
+    [CmdletBinding(HelpUri='http://go.microsoft.com/fwlink/?LinkID=113426', RemotingCapability='None')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This provides a wrapper around Write-Host. In general, we'd like to use Write-Information, but it's not supported on PS 4.0 which we need to support.")]
```
HowardWolosky-MSFT: You should be able to remove the other two `PSAvoidUsingWriteHost` SuppressMessageAttributes now...one is on `Wait-JobWithAnimation` and the other is on `Write-Log`.

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,54 @@ function Send-SBMailMessage
...
+            ValueFromRemainingArguments=$true)]
+        [System.Object] ${Object},
+        [switch] ${NoNewline},
+        [System.Object] ${Separator},
+        [System.ConsoleColor] ${ForegroundColor},
+        [System.ConsoleColor] ${BackgroundColor}
+    )
+
+    # Determine if the host is interactive
+    if ([Environment]::UserInteractive -and ![Bool]([Environment]::GetCommandLineArgs() -like '-noni*') -and (Get-Host).Name -ne 'Default Host')
```
HowardWolosky-MSFT: In a [StackOverflow discussion](http://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode) I saw on this topic, some users were talking about the performance impact of some of these tests.  I'm curious if you notice any change in perceived behavior when running the command in an interactive console window...

StoreBroker/Helpers.ps1
```diff
@@ -442,21 +440,21 @@ function Write-Log
 
         switch ($Level)
         {
-            'Error'   { Write-Error $ConsoleMessage }
```
HowardWolosky-MSFT: Thanks for catching these...

HowardWolosky-MSFT:
It would also be helpful for you could bump the `patch` part of the [version](https://github.com/Microsoft/StoreBroker/blob/master/CONTRIBUTING.md#releasing) as well as part of your change.

StoreBroker/Helpers.ps1
lisaong: they were already removed, I think you may be looking at an older version.

StoreBroker/Helpers.ps1
lisaong: Not sure what you meant by "capitalize CmdletBinding the same way as we do in the other functions."  Do you have an example?

StoreBroker/Helpers.ps1
lisaong: I haven't noticed any difference so far.  It could be because the commands themselves take a while to run (e.g. showing progress).  If there is really a perf impact, we can consider making a follow-up change to cache the state.

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,58 @@ function Send-SBMailMessage
...
+<#
+    .SYNOPSIS
+        Forwards to Write-Host only if the host is interactive, else does nothing.
+
+    .DESCRIPTION
+        A proxy function around Write-Host that detects if the host is interactive
+        before calling Write-Host. Use this instead of Write-Host to avoid failures in
+        non-interactive hosts.
+
+        Boilerplate is generated using these commands:
```
lisaong: Done!

StoreBroker/Helpers.ps1
lisaong: Done!

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,64 @@ function Send-SBMailMessage
...
+    [CmdletBinding(
+        HelpUri='http://go.microsoft.com/fwlink/?LinkID=113426',
+        RemotingCapability='None')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This provides a wrapper around Write-Host. In general, we'd like to use Write-Information, but it's not supported on PS 4.0 which we need to support.")]
+    param(
+        [Parameter(
+            Position=0,
+            ValueFromPipeline=$true,
+            ValueFromRemainingArguments=$true)]
+
```
HowardWolosky-MSFT: nit: No linefeed between the [Parameter] attribute and the parameter that it's associated with,

StoreBroker/Helpers.ps1
```diff
@@ -645,4 +643,64 @@ function Send-SBMailMessage
...
+#>
+
+    [CmdletBinding(
+        HelpUri='http://go.microsoft.com/fwlink/?LinkID=113426',
+        RemotingCapability='None')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "", Justification="This provides a wrapper around Write-Host. In general, we'd like to use Write-Information, but it's not supported on PS 4.0 which we need to support.")]
+    param(
+        [Parameter(
+            Position=0,
+            ValueFromPipeline=$true,
```
HowardWolosky-MSFT: nit: no need to specify `$true` for the parameter configurations.  When you reference them by name, they default to `$true`.

PR #15, opened 2017-01-26
v1.1.1 - Add a logger option to print timestamps in UTC for log traces

### LINKED ISSUE #10: Add a logger option to print timestamps in UTC for log traces
Currently, the logger outputs timestamps (which are super useful for diagnosis), but they default to local time.  

When StoreBroker is used in conjunction with web applications, being able to output UTC timestamps will simplify correlating with other log sources that do the same.

HowardWolosky-MSFT:
`Write-Log` is used for writing all messages to the console and the log file.
It embeds a timestamp with each message, but this timestamp uses LocalTime.
A request has come in to optionally allow users to have this timestamp
recorded in UTC instead of LocalTime.

This adds a new global variable `$global:SBUseUTC` that defaults to `$false`.
If set to `$true`, then `Write-Log` will use UTC for the timestamps.

The documentation has been updated to reflect this new
configuration option.

* Additionally fixed a typo in `Write-Log` where it was trying to
call `Write-Debug`.

Resolves #10: Add a logger option to print timestamps in UTC for log traces

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -71,6 +71,11 @@ function Initialize-HelpersGlobalVariables
     {
         $global:SBNotifyCredential = [PSCredential]$null
     }
+
+    if (!(Get-Variable -Name SBUseUTC -Scope Global -ValueOnly -ErrorAction SilentlyContinue))
+    {
+        $global:SBUseUTC = $false
+    }
 }
 
 # We need to be sure to call this explicitly so that the global variables get initialized.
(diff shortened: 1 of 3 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -415,16 +420,23 @@ function Write-Log
 
     Process
     {
+        $date = Get-Date
+        $dateString = $date.ToString("yyyy-MM-dd HH:mm:ss")
+        if ($global:SBUseUTC)
+        {
+            $dateString = $date.ToUniversalTime().ToString("yyyy-MM-dd HH:mm:ss")
+        }
+
```
danbelcher-MSFT: It looks like both the Local Time and UTC are being presented in the same format?  How can we tell if a given log file uses UTC or Local Time?  Are we logging the value of $global:SBUseUTC?

lisaong: You can also use the 'Z' postfix or offsets for the timestamps: http://www.w3.org/TR/NOTE-datetime

HowardWolosky-MSFT: Good suggestion. Done.

PR #18, opened 2017-02-07
Fix static analysis warnings

TheHoward:
There are some new warnings coming up when checking StoreBroker
with newer versions of [PSScriptAnalyzer](https://github.com/PowerShell/PSScriptAnalyzer).

* `PSUseDeclaredVarsMoreThanAssignments` - We had some instances where
  a variable was being assigned to but never used.  For the API instances,
  we now capture to $null in the instances where we don't want the helper
  method's results being returned to the user.

* `PSUseShouldProcessForStateChangingFunctions` - Two PackageTool methods
  use the verb "Remove" but have no need for providing ShouldProcess support.
  Those instances are now suppressed.

All remaining warnings are false positives of `PSUseDeclaredVarsMoreThanAssignments`.
I have opened two different issues against PSScriptAnalyzer to track these false
positives:
   * [PSUseDeclaredVarsMoreThanAssignment not correctly handling global vars assigned to within functions](PowerShell/PSScriptAnalyzer#698)
   * [PSUseDeclaredVarsMoreThanAssignment not correctly handling array adds](PowerShell/PSScriptAnalyzer#699)

Updated CONTRIBUTING.md to remind users to keep the analyzer module up-to-date.

Additionally, the links in PDP.md were invalid.  Those have now been fixed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ into the source.
 
 StoreBroker is maintained by:
 
-- **[@HowardWolosky-MSFT](http://github.com/HowardWolosky-MSFT)**
+- **[@TheHoward](http://github.com/TheHoward)**
 - **[@DanBelcher-MSFT](http://github.com/DanBelcher-MSFT)**
 
 As StoreBroker is a production dependency for Microsoft, we have a couple workflow restrictions:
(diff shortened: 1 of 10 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -124,27 +125,27 @@ function Wait-JobWithAnimation
     $iteration = 0
     while (((Get-Job -Name $JobName).state -eq 'Running'))
     {
-        Write-InteractiveHost "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
+        Write-Host "`r$($animationFrames[$($iteration % $($animationFrames.Length))])  Elapsed: $([int]($iteration / $framesPerSecond)) second(s) $Description" -NoNewline -f Yellow
         Start-Sleep -Milliseconds ([int](1000/$framesPerSecond))
```
danbelcher-MSFT: Looks like you need to sync with master.  The changes to Write-InteractiveHost are being changed back to Write-Host

TheHoward: Weird.  Good catch, fixed.

PR #23, opened 2017-02-17
Fix proxy request encoding and return special headers

HowardWolosky:
The proxy was submitting all requests with the encoding type `utf8`.
The correct value for that encoding type is `UTF-8`.  The Windows
Store Submission API was robust with the improperly specified encoding
type for some commands, but not others.  This gets the request to be
standard across the board.

Additionally, this updates the proxy to return back all of the special
response headers that the Windows Store Submission API returns.
Most importantly, this is now returning `MS-CorrelationID` which is
used by the Windows Store Submission API team when they are investigating
bug reports.

At this time, StoreBroker itself will never see/report this value because
the PowerShell Invoke-RestMethod (which Invoke-SBRestMethod uses under the
hood) discards all HTTP response headers without any option to prevent
that behavior.  Having these additional headers though will still be useful
if one-off investigations need to be performed.

### DIFF
--- a/RESTProxy/Models/Endpoint.cs
+++ b/RESTProxy/Models/Endpoint.cs
@@ -285,7 +285,7 @@ namespace Microsoft.Windows.Source.StoreBroker.RestProxy.Models
                 {
                     byte[] bytes = System.Text.Encoding.UTF8.GetBytes(body);
                     request.ContentLength = bytes.Length;
-                    request.ContentType = $"{ProxyManager.JsonMediaType}; charset=utf8";
+                    request.ContentType = $"{ProxyManager.JsonMediaType}; charset=UTF-8";
 
                     using (Stream requestStream = await request.GetRequestStreamAsync())
                     {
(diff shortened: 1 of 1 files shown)


RESTProxy/Models/Endpoint.cs
```diff
@@ -285,7 +285,7 @@ public Endpoint Duplicate()
                 {
                     byte[] bytes = System.Text.Encoding.UTF8.GetBytes(body);
                     request.ContentLength = bytes.Length;
-                    request.ContentType = $"{ProxyManager.JsonMediaType}; charset=utf8";
+                    request.ContentType = $"{ProxyManager.JsonMediaType}; charset=UTF-8";
```
danbelcher-MSFT: Are there plans for the Submission API team to make their APIs resilient against this issue?

HowardWolosky: Yes.  They've opened a bug on their end to improve their handling of non-standard encoding values.

PR #24, opened 2017-02-17
Properly escape special characters when generating config files

HowardWolosky:
We were escaping linefeeds when generating config files, but we
were not escaping back slashes, quote characters or tabs.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -279,7 +279,7 @@ function Get-SHA512Hash
     .SYNOPSIS
         Gets the SHA512 hash of the requested string.
 
-    .SYNOPSIS
+    .DESCRIPTION
         Gets the SHA512 hash of the requested string.
 
         The Git repo for this module can be found here: http://aka.ms/StoreBroker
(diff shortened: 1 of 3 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -307,6 +307,50 @@ function Get-SHA512Hash
...
+    [CmdletBinding()]
+    param(
+        [Parameter(
+            Mandatory,
+            ValueFromPipeline)]
+        [AllowNull()]
+        [AllowEmptyString()]
+        [string] $Value
+    )
+
```
danbelcher-MSFT: Would be nice to have a comment acknowledging the confusion for future readers.

StoreBroker/Helpers.ps1
```diff
@@ -307,6 +307,50 @@ function Get-SHA512Hash
...
 }
 
+function Get-EscapedJsonValue
+{
+<#
+    .SYNOPSIS
+        Escapes special characters within a string for use within a JSON value.
+
+    .SYNOPSIS
+        Escapes special characters within a string for use within a JSON value.
```
danbelcher-MSFT: .Description

StoreBroker/Helpers.ps1
```diff
@@ -307,6 +307,50 @@ function Get-SHA512Hash
...
+        Get-EscapedJsonValue -Value 'This is my "quote". Look here: c:\windows\'
+
+        Returns back the string 'This is my \"quote\". Look here: c:\\windows\\'
+
+    .OUTPUTS
+        System.String - A string with special characters escaped for use within JSON.
+
+    .NOTES
+        Normalizes newlines and carraige returns to always be \r\n.
+#>
```
danbelcher-MSFT: typo: carraige

StoreBroker/Helpers.ps1
HowardWolosky: Good catch.  That was from a copy/paste error with `Get-SHA512Hash`, so I've fixed it there too.

StoreBroker/Helpers.ps1
HowardWolosky: Done

PR #25, opened 2017-02-17
Clarify behavior of Dev Portal with new submissions

HowardWolosky:
We have a warning on Commit that changes to a submission can't be seen
in the Dev Portal until the submission enters the Certification phase,
however we weren't providing that warning right after a submission was
patched, which was leading to some user confusion of why they couldn't
see their changes.

The warning has been moved to a single location and is now referenced
throughout the module.

Additionally, normalized the capitalization of the phrase "Dev Portal"
throughout the module.

### DIFF
--- a/StoreBroker/StoreIngestionApplicationApi.ps1
+++ b/StoreBroker/StoreIngestionApplicationApi.ps1
@@ -1106,10 +1106,12 @@ function Update-ApplicationSubmission
 
         $output = @()
         $output += "Successfully cloned the existing submission and modified its content."
-        $output += "You can view it on the Dev portal here:"
+        $output += "You can view it on the Dev Portal here:"
         $output += "    https://dev.windows.com/en-us/dashboard/apps/$AppId/submissions/$submissionId/"
         $output += "or by running this command:"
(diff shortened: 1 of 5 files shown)


StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -1106,10 +1106,12 @@ function Update-ApplicationSubmission
 
         $output = @()
         $output += "Successfully cloned the existing submission and modified its content."
-        $output += "You can view it on the Dev portal here:"
+        $output += "You can view it on the Dev Portal here:"
         $output += "    https://dev.windows.com/en-us/dashboard/apps/$AppId/submissions/$submissionId/"
         $output += "or by running this command:"
         $output += "    Get-ApplicationSubmission -AppId $AppId -SubmissionId $submissionId | Format-ApplicationSubmission"
+        $output += ""
+        $output += $script:manualPublishWarning -f 'Update-ApplicationSubmission'
```
danbelcher-MSFT: Wondering if we should get in the habit of creating string constants for referencing our functions, since we do that a lot.

HowardWolosky: Interesting suggestion. Won't block this change on that suggestion, but will consider it for the future.

PR #30, opened 2017-03-07
Publish a StoreBroker NuGet package.

### LINKED ISSUE #1: Create and publish NuGet package
We'd like to support a simpler installation option of StoreBroker using NuGet (and publish the package through nuget.org)

The package should include:

      \Documentation\*
      \Extensions\*
      \PDP\*
      \StoreBroker\*
      \*.md
      \LICENSE

danbelcher-MSFT:
This change adds support for publishing a StoreBroker NuGet package. The CONTRIBUTING.md is updated to reference the internal repo as the location
for documentation describing the packaging/publishing process.

StoreBroker.nuspec is added to the root directory and describes the content for the .nupkg. The author of the package is Microsoft and the package will be co-owned by the 'microsoft' and 'nugetstorebrokerteam' NuGet accounts.

The .nupkg will contain the entirety of the StoreBroker subfolder (the files constituting the module), both of the Extensions\ConvertFrom-* scripts, and the XSD files from the PDP folder (for validating PDP XML schema).

Resolves #1 : Create and publish NuGet package

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -301,6 +301,10 @@ Where:
 * `<minor>` - If this is a feature update, increment by one and be sure to reset `<patch>` to 0.
 * `<patch>` - If this is a bug fix, leave `<minor>` alone and increment this by one.
 
+When new code changes are checked in to the repo, a new NuGet package must be published by Microsoft.
+This process is documented in Microsoft's internal StoreBroker repo.
+Refer to the CONTRIBUTING.md in that repo for more information on creating a signed NuGet package.
+
 ----------
 
 ### Legal and Licensing
--- a/StoreBroker.nuspec
+++ b/StoreBroker.nuspec
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.Windows.StoreBroker</id>
+    <version>$version$</version>
+    <authors>Microsoft</authors>
+    <owners>microsoft,nugetstorebrokerteam</owners>
+    <licenseUrl>https://aka.ms/StoreBroker_License</licenseUrl>
+    <projectUrl>https://aka.ms/StoreBroker</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</iconUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>A PowerShell module that leverages the Windows Store Submission API to allow easy automation of application submissions to the Windows Store.</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <tags>microsoft windows store broker storebroker submission api powershell posh module</tags>
+    <language>en-US</language>
+  </metadata>
+  <files>
+      <file src="StoreBroker\" target="StoreBroker\"/>
+      <file src="Extensions\" target="Extensions\"/>
+      <file src="PDP\*.xsd" target="PDP\"/>
+  </files>
+</package>
--- a/StoreBroker/NugetTools.ps1
+++ b/StoreBroker/NugetTools.ps1
@@ -30,7 +30,7 @@ function Get-NugetExe
         $sourceNugetExe = "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
         $script:nugetExePath = Join-Path $(New-TemporaryDirectory) "nuget.exe"
     
-        Write-Log "Downlading $sourceNugetExe to $script:nugetExePath" -Level Verbose
+        Write-Log "Downloading $sourceNugetExe to $script:nugetExePath" -Level Verbose
         Invoke-WebRequest $sourceNugetExe -OutFile $script:nugetExePath
     }
 
--- a/StoreBroker/StoreBroker.psd1
+++ b/StoreBroker/StoreBroker.psd1
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.7.0'
+    ModuleVersion = '1.7.1'
     Description = 'Provides command-line access to the Windows Store Submission REST API.'
 
     RootModule = 'StoreIngestionApi'
@@ -143,10 +143,10 @@
             Tags = @('Store', 'App', 'Submission')
 
             # A URL to the license for this module.
-            LicenseUri = 'http://aka.ms/StoreBroker_License'
+            LicenseUri = 'https://aka.ms/StoreBroker_License'
 
             # A URL to the main website for this project.
-            ProjectUri = 'http://aka.ms/StoreBroker'
+            ProjectUri = 'https://aka.ms/StoreBroker'
 
             # A URL to an icon representing this module.
             # IconUri = ''


Documentation/NUGET.md
```diff
@@ -0,0 +1,62 @@
...
+### Sign In
+
+1.  Navigate to [the NuGet website](https://www.nuget.org/).
+2.  Click `Register / Sign in` near the top of the page.
+3.  At the sign-in page, click `Lost your Password`.
+4.  Send the password reset email to `nugetstorebrokerteam@microsoft.com`.
+5.  Open the email (check your spam folder if you do not see it) and follow the reset link.
+6.  The new password will be a GUID. From a PowerShell console, run:
+
+        [Guid]::NewGuid()
```
HowardWolosky: Could possibly suggest this instead:

`(Get-Culture).TextInfo.ToTitleCase([Guid]::NewGuid())`

Extensions/New-NuGetPackage.ps1
```diff
@@ -37,12 +45,95 @@ param(
...
+        $nugetExePath = Join-Path -Path $script:tempFolderPath -ChildPath "nuget.exe"
+        
+        Write-Host "Downloading [" -NoNewline
+        Write-Host "$sourceNugetExe" -ForegroundColor Yellow -NoNewline
+        Write-Host "] to [" -NoNewline
+        Write-Host "$nugetExePath" -ForegroundColor Yellow -NoNewline
+        Write-Host "]."
+    
+        $out = @()
+        $out += "To prevent this download in the future, download NuGet.exe from NuGet.org and"
```
HowardWolosky: instead of saying "download NuGet.exe from NuGet.org and make sure it is accessible from the system PATH", why not just say "copy $nugetExePath to a directory accessible from the system PATH"?

Extensions/New-NuGetPackage.ps1
```diff
@@ -37,12 +45,95 @@ param(
...
+{
+    # Get the Manifest content which contains the module version number.
+    $manifestContent = (Get-Content -Path $ManifestPath) -join [Environment]::NewLine
+
+    # The Manifest file is just a PoSh hashtable.
+    # Convert the string to hashtable.
+    $manifest = Invoke-Expression -Command $manifestContent
+
+    # Create the package using the module version
+    & (Get-NugetExe) pack $NuSpecPath -Properties version=$($manifest.ModuleVersion) -OutputDirectory $OutPath -NoPackageAnalysis
```
HowardWolosky: if `$OutPath` is specified as a relative path, will this still work? Does the path need to be resolved first?

HowardWolosky: Why don't we want package analysis?

Extensions/New-NuGetPackage.ps1
```diff
@@ -37,12 +45,95 @@ param(
...
+}
+finally
+{
+    if (($null -ne $script:tempFolderPath) -and (Test-Path -PathType Container -Path $script:tempFolderPath))
```
HowardWolosky: Add a comment that this is being set in `Get-NugetExe`, and then add a `.NOTES` section to `Get-NugetExe` that a side-effect of running that script is the potential setting of that script variable.  That script variable should also be set to `$null` by default for clarity.

StoreBroker/NugetTools.ps1
```diff
@@ -30,7 +30,7 @@ function Get-NugetExe
         $sourceNugetExe = "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
         $script:nugetExePath = Join-Path $(New-TemporaryDirectory) "nuget.exe"
     
-        Write-Log "Downlading $sourceNugetExe to $script:nugetExePath" -Level Verbose
+        Write-Log "Downloading $sourceNugetExe to $script:nugetExePath" -Level Verbose
```
HowardWolosky: Thanks for the fix.  Since you are changing a file in the module though, increment the patch number in the module manifest.

Extensions/New-NuGetPackage.ps1
danbelcher-MSFT: I clean up the temporary directory when the script finishes, in order to not clutter the computer, so the downloaded executable no longer exists.  Do you prefer that I leave the temp folder and print this message you suggested?

Extensions/New-NuGetPackage.ps1
danbelcher-MSFT: Yes, still works with a relative path

danbelcher-MSFT: The package analysis dumps a bunch of warnings, none of which are useful, and recommends an action that we don't want.  Every .ps1 gets a 3-line warning: 
WARNING: Issue: PowerShell file outside tools folder.
WARNING: Description: The script file 'StoreBrokerNuGet\StoreBroker\NugetTools.ps1' is outside the 'tools' folder and hence will not be executed during installation of this package.
WARNING: Solution: Move it into the 'tools' folder.

Extensions/New-NuGetPackage.ps1
HowardWolosky: Right, right.  I wrote that comment before I saw your other code later.  Leave what you have as-is.

Extensions/New-NuGetPackage.ps1
HowardWolosky: Hmmm...tried doing some searching to see if there's something that can be done to mark those as ignorable and still get the analysis, but came up short.  It's probably worth a comment there to indicate that analysis is skipped because it gives false-positive warnings since we're not putting our PowerShell scripts into the Tools directory.

Extensions/New-NuGetPackage.ps1
```diff
@@ -0,0 +1,139 @@
...
+        while (Test-Path -PathType Container -Path (Join-Path -Path $env:TEMP -ChildPath $guid))
+
+        $script:tempFolderPath = Join-Path -Path $env:TEMP -ChildPath $guid
+        Write-Host "Creating temporary directory: $script:tempFolderPath"
+        New-Item -ItemType directory -Path $script:tempFolderPath | Out-Null
+    
+        $sourceNugetExe = "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+        $nugetExePath = Join-Path -Path $script:tempFolderPath -ChildPath "nuget.exe"
+        
+        Write-Host "Downloading [" -NoNewline
```
HowardWolosky: This seems a bit unnecessary.  I tried it and didn't get a lot of value out of it, visually.
Just stick with a regular, old-fashioned single line `Write-Host` without bothering with `ForegroundColor`.  It'll make the code easier t read that way too.  Skip the coloring on the next one too.

danbelcher-MSFT: Removed

StoreBroker/StoreBroker.psd1
```diff
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.4.5'
+    ModuleVersion = '1.4.6'
```
danbelcher-MSFT: Will update ModuleVersion appropriately when it's time to check in

Extensions/New-NuGetPackage.ps1
```diff
@@ -0,0 +1,214 @@
...
+                throw ($out -join [Environment]::NewLine)
+            }
+        }
+    }
+}
+
+# New-NuGetPackage.ps1 body
+try
+{
+    if (-not $IgnoreUnsigned.IsPresent)
```
HowardWolosky: Interesting.  I wasn't aware of the `IsPresent` property on switches.  I think that could be useful in the future.

Documentation/NUGET.md
```diff
@@ -0,0 +1,78 @@
...
+6.  Use a GUID for the new password. From a PowerShell console, run:
+
+        [Guid]::NewGuid()
+
+    Take the output and capitalize the first letter in the GUID, in order to satisfy NuGet's password
+    requirements. Use the result as your new password. Do not store or save this password anywhere.
+    Every time you need to access the account you should follow the password reset process.
+7.  Sign in using the new password.
+
+### Create an API Key
```
HowardWolosky: Will you be updating this with publishing steps as well?

Documentation/NUGET.md
```diff
@@ -0,0 +1,78 @@
...
+is restricted to Microsoft employees.
+
+----------
+
+
+## Process
+
+This document describes the process for releasing an updated version of the StoreBroker NuGet package.
+
+### Signing
```
HowardWolosky: This should note that the signed files should be pushed into master as well.

CONTRIBUTING.md
```diff
@@ -301,6 +301,9 @@ Where:
 * `<minor>` - If this is a feature update, increment by one and be sure to reset `<patch>` to 0.
 * `<patch>` - If this is a bug fix, leave `<minor>` alone and increment this by one.
 
+When new code changes are checked in to the repo, a new NuGet package must be published. This process is documented in
```
HowardWolosky: must be published `by Microsoft`.

CONTRIBUTING.md
```diff
@@ -301,6 +301,9 @@ Where:
 * `<minor>` - If this is a feature update, increment by one and be sure to reset `<patch>` to 0.
 * `<patch>` - If this is a bug fix, leave `<minor>` alone and increment this by one.
 
+When new code changes are checked in to the repo, a new NuGet package must be published. This process is documented in
+the internal repo. Refer to the CONTRIBUTING.md in that repo for more information on creating a signed NuGet package.
```
HowardWolosky: documented in `Microsoft's` internal `StoreBroker` repo.

StoreBroker.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288890</iconUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>A PowerShell module that leverages the Windows Store Submission API to allow easy automation of application submissions to the Windows Store.</description>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <tags>microsoft windows store broker storebroker submission api powershell posh module</tags>
+    <language>en-US</language>
+  </metadata>
+  <files>
+      <file src="StoreBroker\" target="StoreBroker\"/>
+      <file src="Extensions\ConvertFrom-ExistingSubmission.ps1" target="Extensions\"/>
```
HowardWolosky: Curious ... why call out these individual files here as opposed to just having _all_ extensions go to `Extensions` the way that you do for `StoreBroker\`?

StoreBroker/StoreBroker.psd1
```diff
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.6.1'
+    ModuleVersion = '1.7.0'
```
HowardWolosky: You'll need to update this again.  Just update the patch version though (so `1.7.1`).

StoreBroker.nuspec
danbelcher-MSFT: Ah, this was because an earlier iteration introduced a script here that we didn't want packaged.  That script is moved to the internal repo so it will be fine to just have the Extensions folder.  Good feedback.

PR #34, opened 2017-03-09
Enable relative package paths for uploading/downloading

HowardWolosky:
PowerShell-native commands can work just fine with relative file
paths, however when interacting with other code, it's necessary
to use `Convert-Path` to first resolve the path to its full path.

`Set-SubmissionPackage` / `Get-SubmissionPackage` use the Azure
assemblies for uploading/downloading packages, and thus need to
have `$PackagePath` resolved to its full path in the even that
users are providing relative paths to the command.

Additionally, fixing the telemetry bucket name used
for `Get-SubmissionPackage`.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -756,3 +756,60 @@ function Write-InteractiveHost
         Write-Host @PSBoundParameters
     }
 }
+
+function Resolve-UnverifiedPath
+{
+<#
+    .SYNOPSIS
+        A wrapper around Resolve-Path that works for paths that exist as well
+        as for paths that don't (Resolve-Path normally throws an exception if
+        the path doesn't exist.)
+
+    .DESCRIPTION
+        A wrapper around Resolve-Path that works for paths that exist as well
+        as for paths that don't (Resolve-Path normally throws an exception if
+        the path doesn't exist.)
+
+        The Git repo for this module can be found here: https://aka.ms/StoreBroker
+
+    .EXAMPLE
+        Resolve-UnverifiedPath -Path 'c:\windows\notepad.exe'
+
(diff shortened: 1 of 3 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -859,6 +862,9 @@ function Get-SubmissionPackage
         [switch] $NoStatus
     )
 
+    # Let's convert this path to a full path so that it works with non-PowerShell commands (like the Azure module)
+    $PackagePath = Convert-Path -Path $PackagePath
```
danbelcher-MSFT: There's a ValidateScript on $PackagePath that specifically checks to make sure the path *does not* exist.  Then you pass that path to Convert-Path which throws an exception for paths that do not exist.  Not sure how this will run correctly.

HowardWolosky: <sigh> That's a good catch.  I tested the change to `Set-SubmissionPackage` but didn't look at the change to `Get-SubmissionPackage` closely enough.

StoreBroker/Helpers.ps1
```diff
@@ -756,3 +756,60 @@ function Write-InteractiveHost
...
+function Resolve-UnverifiedPath
+{
+<#
+    .SYNOPSIS
+        A wrapper around Resolve-Path that works for paths that exist as well
+        as for paths that don't (Resolve-Path normally throws an exception if
+        the path doesn't exist.)
+
+    .DESCRIPTION
+        A wrapper around Resolve-Path that works for paths that exist as well
```
danbelcher-MSFT: typo: 'fdr' -> 'for'

danbelcher-MSFT: Nevermind, something on my monitor XD

PR #39, opened 2017-03-14
Additional package metadata

### LINKED ISSUE #16: Add package metadata to submission JSON
Customer has requested that `New-SubmissionPackage` adds some additional metadata on the packages that it processes to the json that it creates.  All of this information (except for Min OS Version) is already gathered and embedded into the new name for the APPX -- this is just about making that same data more easily consumable via JSON.

Specifically:
- Minimum Target OS Version
- Array of targeted device types (e.g. Mobile, Desktop, Team, etc...)
- Version
- Array of targeted architectures.

This information will then need to be stripped out by `Patch-ApplicationSubmission` since those properties will be invalid with API interaction.

danbelcher-MSFT:
This change adds additional package metadata to the output .JSON from the New-SubmissionPackage command.

New metadata is added to each object in 'applicationPackages'.  The new fields are:
-version
-architecture
-targetPlatform
-languages
-capabilities
-targetDeviceFamilies
-targetDeviceFamiliesEx
-minOSVersion
-innerPackages

These properties are removed during a submission because they are re-calculated by the Store.

Resolves #16

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -24,6 +24,19 @@ $script:s_OutPath = "OutPath"
 $script:s_OutName = "OutName"
 $script:s_DisableAutoPackageNameFormatting = "DisableAutoPackageNameFormatting"
 
+# String constants for application metadata
+$script:applicationMetadataProperties = @(
+    "version",
+    "architecture",
+    "targetPlatform",
+    "languages",
+    "capabilities",
+    "targetDeviceFamilies",
+    "targetDeviceFamiliesEx",
+    "minOSVersion",
+    "innerPackages"
+)
+
 function Get-StoreBrokerConfigFileContentForIapId
 {
 <#
@@ -35,7 +48,7 @@ function Get-StoreBrokerConfigFileContentForIapId
         Updates the default IAP configuration file template with the values from the
         indicated IAP's most recent submission.
 
-        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+        The Git repo for this module can be found here: https://aka.ms/StoreBroker
 
     .PARAMETER ConfigContent
         The content of the config file template as a simple string.
@@ -132,7 +145,7 @@ function New-StoreBrokerInAppProductConfigFile
         Creates a new configuration file as a template for an In-App Product submission.
         The full path to the new file can be provided by the -Path parameter.
 
-        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+        The Git repo for this module can be found here: https://aka.ms/StoreBroker
 
     .PARAMETER Path
         A full path specifying where the new config file will go and what it will be
@@ -207,7 +220,7 @@ function Get-StoreBrokerConfigFileContentForAppId
         Updates the default configuration file template with the values from the
         indicated App's most recent published submission.
 
-        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+        The Git repo for this module can be found here: https://aka.ms/StoreBroker
 
     .PARAMETER ConfigContent
         The content of the config file template as a simple string.
@@ -325,7 +338,7 @@ function New-StoreBrokerConfigFile
         Creates a new configuration file as a template for an app submission.
         The full path to the new file can be provided by the -Path parameter.
 
-        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+        The Git repo for this module can be found here: https://aka.ms/StoreBroker
 
     .PARAMETER Path
         A full path specifying where the new config file will go and what it will be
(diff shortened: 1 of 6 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+        $metadata.targetDeviceFamilies = @()
+        foreach ($family in $metadata.targetDeviceFamiliesEx)
+        {
+            $metadata.targetDeviceFamilies += ("{0} min version {1}" -f $family, $metadata.minOSVersion)
+        }
+
+        # Track the info about this package for later processing.
+        $singleAppxInfo = @{}
+        $singleAppxInfo[[StoreBrokerTelemetryProperty]::AppxVersion] = $metadata.version
+        $singleAppxInfo[[StoreBrokerTelemetryProperty]::AppName] = $metadata.name -split "Microsoft.", 0, "SimpleMatch" |
```
HowardWolosky: I thought $metadata.name already had that `-split` logic done to it above.  Why do you need to do it again?

StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+                languages            = $appxMetadata.languages;
+                capabilities         = $appxMetadata.capabilities;
+                minOSVersion         = $appxMetadata.minOSVersion;
+                targetPlatform       = $appxMetadata.targetPlatform;
             }
+        }
 
-            return $formattedIdentity
+        # Guarantee uniqueness
+        $metadata.capabilities           += $capabilities | Sort-Object -Unique
```
HowardWolosky: Are you sure this guarantees uniqueness?  You're doing a `+=` to append unique data, but not guaranteeing that the data being added is unique to what's already in the array that you're appending to.  Same for below.

StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+    $version = $Metadata.version
+    if ($null -ne $Metadata.innerPackages)
+    {
+        # For .appxbundle packages, we will use the architectures from the individual .appx files.
+        # The Keys of the innerPackages object are the supported architectures.
+        $architectureTag = ($Metadata.innerPackages.Keys | Sort-Object) -join '.'
+
+        # Grab an arbitrary one and use that version.
+        $arch = $Metadata.innerPackages.Keys | Sort-Object | Select -First 1
+        $version = $Metadata.innerPackages.$arch.version
```
HowardWolosky: I was thinking maybe we should change to using the bundle version in this scenario.  Can you confirm with Release that they would be ok with that?

StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+    }
+
+    # Simplify 'Windows.Universal' to 'Universal'
+    $target = "Windows."
+    $deviceFamilyCollection = @()
+    foreach ($family in $Metadata.targetDeviceFamiliesEx)
+    {
+        $index = $family.IndexOf($target)
```
HowardWolosky: I think this section could be simplified significantly to just do this:
`$target -replace '^Windows.', ''`
(that says, if the string starts with "Windows.", replace that with nothing.  Then return the resulting string regardless.

StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+        }
+        
+        $deviceFamilyCollection += $family
+    }
+
+    # Categories are joined with a '_' separator
+    $formattedBundleTags = @($Metadata.name, $version, $architectureTag)
+    if ($deviceFamilyCollection.Count -gt 0)
+    {
+        $formattedBundleTags = @(($deviceFamilyCollection | Sort-Object) -join '.') + $formattedBundleTags
```
HowardWolosky: I feel like this line would read easier as:
`$formattedBundleTags += ($deviceFamilyCollection | Sort-Object) -join '.')`

StoreBroker/PackageTool.ps1
```diff
@@ -1294,151 +1324,565 @@ function Report-UnsupportedFile
...
+                $metadata = Read-AppxUploadMetadata -AppxuploadPath $AppxPath -AppxInfo $AppxInfo
+            }
+
+            ".appx"
+            {
+                $metadata = Read-AppxMetadata -AppxPath $AppxPath -AppxInfo $AppxInfo
             }
         }
+
+        $metadata.name = Get-FormattedFilename -Metadata $metadata
```
HowardWolosky: I'm confused that the inner methods are calculating `$metadata.name`, but then you're replacing it here.  Should they be storing their property somewhere else?  Maybe use a different property name here like `formattedFileName`?

StoreBroker/PackageTool.ps1
```diff
@@ -1509,27 +1953,32 @@ function Add-AppPackagesMetadata
...
-                fileStatus = "PendingUpload";
-                version = $null;
-                architecture = $null;
-                languages = [System.Array]::CreateInstance([Object], 0);
-                capabilities = [System.Array]::CreateInstance([Object], 0);
-                minimumDirectXVersion = "None";
-                minimumSystemRam = "None";
+            # Finalize the properties to be submitted
+            $submissionProperties = @{}
+            foreach ($property in $script:applicationMetadataProperties)
```
HowardWolosky: Ah...now having read through everything, I see that you're adding additional properties (like `name` and `formattedFilename`) to here that we wouldn't add.  I guess you can avoid this comment and my earlier one about moving that array to be a member var of the `New-*` function

StoreBroker/Helpers.ps1
```diff
@@ -813,3 +813,52 @@ function Resolve-UnverifiedPath
         return $resolvedPath.ProviderPath
     }
 }
+
+function Remove-UnofficialSubmissionProperties
```
HowardWolosky: Nice abstraction

StoreBroker/PackageTool.ps1
danbelcher-MSFT: :)

StoreBroker/Helpers.ps1
danbelcher-MSFT: Moved to StoreIngestionApi.psm1

StoreBroker/PackageTool.ps1
HowardWolosky: Keep as is.  You're totally right.  Misread the code.

StoreBroker/PackageTool.ps1
HowardWolosky: Alrighty.  Sounds good.  Just wanted to make sure we checked with them first.

StoreBroker/PackageTool.ps1
HowardWolosky: Ah, I see that now.  So, to clarify, the `+=` being used is simply to ensure that the property always remains an array?  If so, can you just add a comment below `# guarantees uniqueness` that simply says something to the effect `# Even though these properties are empty up until this point, we use += to ensure that they remain arrays even if there's a single object that we'd be assigning to it`....it would help avoid confusion for future readers of the code.

StoreBroker/PackageTool.ps1
```diff
@@ -24,6 +24,36 @@ $script:s_OutPath = "OutPath"
...
+    "languages",
+    "capabilities",
+    "targetDeviceFamilies",
+    "targetDeviceFamiliesEx",
+    "minOSVersion",
+    "innerPackages"
+)
+
+# String constants for UAP namespaces
+$script:uapNamespaces = @(
```
danbelcher-MSFT: Will change to the string search.  Agreed that it is more future-proof.

StoreBroker/PackageTool.ps1
danbelcher-MSFT: Agreed to keep as is.

StoreBroker/PackageTool.ps1
danbelcher-MSFT: Agreed to use the string search instead of maintaining the namespace set.

StoreBroker/PackageTool.ps1
```diff
@@ -442,7 +425,7 @@ function Write-SubmissionRequestBody
 
     if ($PSCmdlet.ShouldProcess($OutFilePath, "Output to File"))
     {
-        $JsonObject | ConvertTo-Json -Depth $script:jsonConversionDepth -Compress:$false | Out-File -Encoding utf8 -FilePath $OutFilePath
```
HowardWolosky: Curious why you decided to change this

HowardWolosky:
Do a rebase to flatten your change out to a single change, and update the version in the psd1 (increment the minor version (`major.minor.patch`) and reset patch to 0).

PR #41, opened 2017-03-28
Support packaging .appxupload files containing a single .appxbundle.

danbelcher-MSFT:
StoreBroker used to support this scenario but dropped support after documentation indicated an .appxupload should only ever contain a single .appx file, see https://docs.microsoft.com/en-us/windows/uwp/packaging/create-app-package-with-makeappx-tool.

Regardless of the documentation, there are StoreBroker users who have been uploading with this type of file and the Store seems to support it.  This change adds back support for this type of file.

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -1149,8 +1149,7 @@ function Report-UnsupportedFile
 {
 <#
     .SYNOPSIS
-        When we discover an invalid .appx, .appxbundle, or .appxupload,
-        raise a new telemetry event and report the file path.
+        When we discover an invalid file, raise a new telemetry event.
 
     .PARAMETER Path
         Filepath of the file that could not be identified.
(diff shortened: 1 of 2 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1166,9 +1165,6 @@ function Report-UnsupportedFile
     {
         $telemetryProperties = @{ [StoreBrokerTelemetryProperty]::SourceFilePath = (Get-PiiSafeString -PlainText $Path) }
         Set-TelemetryEvent -EventName New-SubmissionPackage-UnsupportedFile -Properties $telemetryProperties
-
-        $type = [System.IO.Path]::GetExtension($Path)
-        Write-Log "Unable to find an $type file in: `"$Path`"" -Level Error
```
danbelcher-MSFT: I decided to remove this comment because it doesn't make any sense..

"Unable to find an .appxupload file in `"S:\my\test.appxupload`"" 

Everywhere Report-UnsportedFile is used, it is followed by a throw with a specific error message which was more helpful.

danbelcher-MSFT: Alternatively, I could pass the error message to this function and have the function Write-Log -Level Error then throw.

HowardWolosky: I think it would look a bit confusing to be calling a method to have it throw, unless you were to rename the method to `Report-UnsupportedFileAndThrow` (so that it's clear when seeing it being used that a throw will be occurring).  But given that there isn't any uniformity in the error messages that are displayed in those scenarios, we don't gain anything by combining the throw into that method in that way.  So, just remove the throw and let the callers do their own throwing.

PR #42, opened 2017-04-04
Add Open-Store

HowardWolosky:
We already had an `Open-DevPortal` function which opened the specified
app in the Developer Portal.  This adds a similar function for opening
the specified app in either the Windows Store web site or in the native
Windows App Store.

### DIFF
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -1347,6 +1347,62 @@ function Open-DevPortal
     }
 }
 
+function Open-Store()
+{
+<#
+.SYNOPSIS
+    Opens the specified app in the Windows Store.
+
+.DESCRIPTION
+    Opens the specified app in the Windows Store.
+
+    The Git repo for this module can be found here: https://aka.ms/StoreBroker
+
+.PARAMETER AppId
+    The ID of the app that should be opened in the Store.
+
+.PARAMETER Web
+    If specified, opens the Web Store instead of the native Windows Store App.
+
+.EXAMPLE
+    Open-Store -AppId 0ABCDEF12345
+
+    Opens the Windows Store app and navigates to the specified application.
+
+.EXAMPLE
+    Open-Store -AppId 0ABCDEF12345 -Web
+
+    Opens the user's browser to the specified app's Windows Store page.
+#>
+    [cmdletbinding()]
+    param(
+        [Parameter(Mandatory)]
+        [string] $AppId,
+
+        [switch] $Web
+    )
+
+    # Telemetry-related
+    $telemetryProperties = @{
+        [StoreBrokerTelemetryProperty]::AppId = $AppId
+        [StoreBrokerTelemetryProperty]::Web = $Web
+    }
+
+    Set-TelemetryEvent -EventName Open-Store -Properties $telemetryProperties
+
+    $webUri = "https://www.microsoft.com/store/apps/$AppId"
+    $storeAppUri = "ms-windows-store://pdp/?productid=$AppId"
+
+    $uri = $storeAppUri
+    if ($Web)
+    {
+        $uri = $webUri
+    }
+
+    Write-Log "Launching $uri" -Level Verbose
+    Start-Process -FilePath $uri
+}
+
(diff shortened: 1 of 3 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1347,6 +1347,61 @@ function Open-DevPortal
...
+    Opens the Windows Store app and navigates to the specified application.
+
+.EXAMPLE
+    Open-Store -AppId 0ABCDEF12345 -Web
+
+    Opens the user's browser to the specified app's Windows Store page.
+#>
+    [cmdletbinding()]
+    param(
+        [Parameter(Mandatory = $true)]
```
danbelcher-MSFT:
```
[Parameter(Mandatory)]
```

Is consistent with the repo

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1347,6 +1347,61 @@ function Open-DevPortal
...
+        [string] $AppId,
+
+        [switch] $Web
+    )
+
+    # Telemetry-related
+    $telemetryProperties = @{
+        [StoreBrokerTelemetryProperty]::AppId = $AppId
+        [StoreBrokerTelemetryProperty]::Web = $Web
```
danbelcher-MSFT: The Web property needs to be defined in Telemetry.ps1

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1347,6 +1347,61 @@ function Open-DevPortal
...
+    Set-TelemetryEvent -EventName Open-Store -Properties $telemetryProperties
+
+    $webUri = "https://www.microsoft.com/store/apps/$AppId"
+    $storeAppUri = "ms-windows-store://pdp/?productid=$AppId"
+
+    $uri = $storeAppUri
+    if ($Web) {
+        $uri = $webUri
+    }
+
```
danbelcher-MSFT: I think it would be just as clear to have
```
if ($Web)
{
    $uri = "..."
}
else
{
    $uri = "..."
}
```
Your call.

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1347,6 +1347,61 @@ function Open-DevPortal
...
+        [StoreBrokerTelemetryProperty]::Web = $Web
+    }
+
+    Set-TelemetryEvent -EventName Open-Store -Properties $telemetryProperties
+
+    $webUri = "https://www.microsoft.com/store/apps/$AppId"
+    $storeAppUri = "ms-windows-store://pdp/?productid=$AppId"
+
+    $uri = $storeAppUri
+    if ($Web) {
```
danbelcher-MSFT: nit: put open curly on new line

StoreBroker/StoreIngestionApi.psm1
danbelcher-MSFT: The Web telemetry property needs to be defined in Telemetry.ps1

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Hmmm... VS Code did that to me.  It originally was.  Thanks,

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: good catch.  I added the telemetry bit right before sending out the PR and hadn't tested it.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Agreed.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: `PSScriptAnalyzer` prefers declared variables, so I need to have an initial value for `$uri`.  I'd rather set it to something meaningful rather than `$null`, so the easiest way is to give it a default valid value.  I'd rather give it a "named" string rather than an anonymous string, and thus, we're at the current implementation.

Issue #43, opened 2017-04-11
Start-SubmissionMonitor should allow the shell to know the result


taoyouh:
The Start-SubmissionMonitor should allow the shell to know the result by returning a status variable, which will be very useful in automated build servers.

HowardWolosky:
@taoyouh - Thanks for the suggestion.  What specifically were you hoping the return value to be?  You're suggesting just the status variable...what if it returned the entire submission so that the status details (like errors/warnings/cert reports) could be examined by your automated build server as well?  Would that still fit your needs?

taoyouh:
@HowardWolosky I was hoping for the last state the submission ends up into. First, I want to have a variable or something that enables me to know if the submission succeeded. Second, I think it will be better if I can know what the errors are.

HowardWolosky:
@taoyouh - Thanks.  Returning the submission object as the result would achieve that, since you could grab the submission status from within it....it would be equivalent to calling Get-ApplicationSubmission afterwards to get the status, since that's what the monitor is doing internally to check on it.

HowardWolosky:
@taoyouh - Please review [PR 45](https://github.com/Microsoft/StoreBroker/pull/45) and give that branch a try to see if it solves your issue.  To grab the final status, you'd do something like:

`$submission = Start-SubmissionMonitor ... 
 $finalStatus = $submission.status`

HowardWolosky:
Fixed in 1.8.1.

PR #45, opened 2017-04-24
Start-SubmissionMonitor should return the final submission retrieved

HowardWolosky:
In order to resolve:
   Issue #43 - Start-SubmissionMonitor should allow the shell to know the result

We now return the final submission object that was retrieved as part
of the loop. Final submission state can be retrieved from that
submission object if desired, or it can be cast to `$null` if not
needed.

### DIFF
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -1128,6 +1128,16 @@ function Start-SubmissionMonitor
         with no commandline status update.  When not specified, those commands run in
         the background, enabling the command prompt to provide status information.
 
+    .PARAMETER PassThru
+        Returns the final submission object that was retrieved when checking submission
+        status.  By default, this function does not generate any output.
+
+    .OUTPUTS
+       None or PSCustomObject
+       By default, this does not generate any output. If you use the PassThru parameter,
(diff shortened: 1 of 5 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1253,6 +1259,8 @@ function Start-SubmissionMonitor
             Start-Sleep -Seconds $secondsBetweenChecks
         }
     }
+
+    return $submission
```
danbelcher-MSFT: What about only returning the object if it's not null?  Having functions output $null can be confusing to debug.  You could end up with the $null as part of an array, for example.

danbelcher-MSFT: In the interest of not breaking any scripts that might depend on this function *not* outputting anything, I think we could add this new behavior with a -PassThru switch.  Any existing scripts won't have the switch so they will be unaffected.  The user requesting this behavior can access it with the new switch.

HowardWolosky: I like this actually. It meshes well with [existing PowerShell conventions](https://blogs.technet.microsoft.com/heyscriptingguy/2011/11/18/use-the-powershell-passthru-parameter-and-get-back-objects/).

Issue #47, opened 2017-04-28
Add support for faster uploads using Azure Storage Data Movement Library


HowardWolosky:
More info on the library can be found [on the GitHub page](https://github.com/Azure/azure-storage-net-data-movement).

We'll need to add a new NuGet dependency (`Microsoft.Azure.Storage.DataMovement`) which builds on top of the pre-existing dependency on `WindowsAzure.Storage`.  The NuGet package brings in a number of other dependencies too, and I'm not yet sure if the rest can be ignored for our scenarios.  More investigation will need to happen.

Right now, in `Set-SubmissionPackage`, after we create the `$cloudBlockBlob` we call the member method `UploadFromFile()`.  It looks like with Azure Storage DML, we'd instead call [`TransferManager.UploadAsync`
](https://azure.github.io/azure-storage-net-data-movement/html/492dd5b3-349b-1483-278c-77b3eab33386.htm).  Calling async methods from PowerShell is not typical, but initial testing indicates we'd be able to call it by calling `[Microsoft.WindowsAzure.Storage.DataMovement.TransferManager]::UploadAsync(...).GetAwaiter().GetResult();` which would turn it into a synchronous call.

We'd also probably want to follow some of the [DML best practices](https://github.com/Azure/azure-storage-net-data-movement#best-practice) around increasing the default connection limit and turning off 100 continue.

If that works, that would retain existing behavior where StoreBroker just sits there until the upload completes.  If we want to do one better, we could look into what it would take to display true upload status (percentage uploaded/remaining as opposed to just time elapsed), by leveraging `SingleTransferContext` and registering a `ProgressHandler` to report back status as the upload occurs.

PR #48, opened 2017-05-01
Add support for faster uploads using Azure Storage Data Movement Library

HowardWolosky:
This update provides about a 10-15x increase in upload performance by
switching from using the standard Windows Azure Storage API's for
uploading/downloading, to the Azure Storage Data Movement API set,
which is optimized for faster uploads.

That change by itself cut the transfer speeds in half.  This then applies
two other suggested best practices for increasing transfer bandwidth:

  * Increasing the `DefaultConnectionLimit`
      > By default, the .Net HTTP connection limit is 2. This implies that
      > only two concurrent connections can be maintained. It prevents more
      > parallel connections accessing Azure blob storage from your application.

  * Turning off `Expect100Continue`
      > When the property "Expect100Continue" is set to true, client requests
      > that use the PUT and POST methods will add an Expect: 100-continue
      > header to the request and it will expect to receive a 100-Continue
      > response from the server to indicate that the client should send the
      > data to be posted. This mechanism allows clients to avoid sending large
      > amounts of data over the network when the server, based on the request
      > headers, intends to reject the request.
      >
      > However, once the entire payload is received on the server end, other
      > errors may still occur. And if Windows Azure clients have tested the
      > client well enough to ensure that it is not sending any bad requests,
      > clients could turn off 100-continue so that the entire request is sent
      > in one roundtrip. This is especially true when clients send small size
      > storage objects.

Some validation tests were performed with sample sizes of 10:

**Upload of a one-gigabye file**
_Current method_
  * Avg: 205 seconds
  * Median: 196 seconds

_Using DataMovement alone_
  * Avg: 98.5 seconds
  * Median: 100 seconds

_Using DataMovement with best practices_
  * Avg: 18 seconds
  * Median: 18 seconds

**Upload of a six-gigabye file**
_Current method_
  * Avg: 1367 seconds
  * Median: 1413 seconds
_*Note: This took so much longer that I only tried two samples._

_Using DataMovement with best practices_
  * Avg: 90.6 seconds
  * Median: 91.5 seconds

Because these results were so surprising, I then downloaded the file after uploading it,
and used `fc.exe /b <orig file> <downloaded file>` to validate the file was indeed uploaded
successfully and completely.

Resolves Issue #47: Add support for faster uploads using Azure Storage Data Movement Library

### DIFF
--- a/StoreBroker/NugetTools.ps1
+++ b/StoreBroker/NugetTools.ps1
@@ -146,6 +146,72 @@ function Get-NugetPackage
     }
 }
 
+function Test-AssemblyIsDesiredVersion
+{
+    <#
+    .SYNOPSIS
+        Checks if the specified file is the expected version.
+
+    .DESCRIPTION
+        Checks if the specified file is the expected version.
+
+        Does a best effort match.  If you only specify a desired version of "6",
+        any version of the file that has a "major" version of 6 will be considered
+        a match, where we use the terminology of a version being:
+        Major.Minor.Build.PrivateInfo.
+
+        The Git repo for this module can be found here: http://aka.ms/StoreBroker
+
+    .PARAMETER AssemblyPath
+        The full path to the assembly file being tested.
+
+    .PARAMETER DesiredVersion
+        The desired version of the assembly.  Specify the version as specifically as
+        necessary.
+
+    .EXAMPLE
+        Test-AssemblyIsDesiredVersion "c:\Microsoft.WindowsAzure.Storage.dll" "6"
+
+        Returns back $true if "c:\Microsoft.WindowsAzure.Storage.dll" has a major version
+        of 6, regardless of its Minor, Build or PrivateInfo numbers.
+
+    .OUTPUTS
+        Boolean - $true if the assembly at the specified path exists and meets the specified
+        version criteria, $false otherwise.
+#>
+    param(
+        [Parameter(Mandatory)]
+        [ValidateScript( { if (Test-Path -PathType Leaf -Path $_) { $true }  else { throw "'$_' cannot be found." } })]
+        [string] $AssemblyPath,
+
+        [Parameter(Mandatory)]
+        [ValidateScript( { if ($_ -match '^\d+(\.\d+){0,3}$') { $true } else { throw "'$_' not a valid version format." } })]
+        [string] $DesiredVersion
+    )
+
+    $splitTargetVer = $DesiredVersion.Split('.')
+
+    $versionInfo = (Get-Item -Path $AssemblyPath).VersionInfo
+    $splitSourceVer = @(
+        $versionInfo.ProductMajorPart,
+        $versionInfo.ProductMinorPart,
+        $versionInfo.ProductBuildPart,
+        $versionInfo.ProductPrivatePart
+    )
+
+    # The cmdlet contract states that we only care about matching
+    # as much of the version number as the user has supplied.
+    for ($i = 0; $i -lt $splitTargetVer.Count; $i++)
+    {
+        if ($splitSourceVer[$i] -ne $splitTargetVer[$i])
+        {
+            return $false
+        }
+    }
+
+    return $true
+}
+
 function Get-NugetPackageDllPath
 {
 <#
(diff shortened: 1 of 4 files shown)


StoreBroker/NugetTools.ps1
```diff
@@ -146,6 +146,79 @@ function Get-NugetPackage
...
+        Boolean - $true if the assembly at the specified path exists and meets the specified
+        version criteria, $false otherwise.
+#>
+    param(
+        [Parameter(Mandatory)]
+        [ValidateScript( { if (Test-Path -PathType Leaf -Path $_) { $true }  else { throw "'$_' cannot be found." } })]
+        [string] $AssemblyPath,
+
+        [Parameter(Mandatory)]
+        [string] $DesiredVersion,
```
danbelcher-MSFT: Don't know if you care to add this, but we could do a regex match to make sure the $DesiredVersion is formatted correctly

StoreBroker/NugetTools.ps1
```diff
@@ -146,6 +146,79 @@ function Get-NugetPackage
...
+
+        [switch] $TreatAsMinimum
+    )
+
+    $versionInfo = (Get-Item $AssemblyPath).VersionInfo
```
danbelcher-MSFT: Named parameter

StoreBroker/NugetTools.ps1
```diff
@@ -146,6 +146,79 @@ function Get-NugetPackage
...
+#>
+    param(
+        [Parameter(Mandatory)]
+        [ValidateScript( { if (Test-Path -PathType Leaf -Path $_) { $true }  else { throw "'$_' cannot be found." } })]
+        [string] $AssemblyPath,
+
+        [Parameter(Mandatory)]
+        [string] $DesiredVersion,
+
+        [switch] $TreatAsMinimum
```
danbelcher-MSFT: I don't think anything uses this switch?  What do we need this for?

StoreBroker/NugetTools.ps1
```diff
@@ -230,8 +303,15 @@ function Get-NugetPackageDllPath
...
-        Write-Log "Found $AssemblyName in module directory ($PSScriptRoot)." -Level Verbose
-        return $moduleAssembly
+        if (Test-AssemblyIsDesiredVersion -AssemblyPath $moduleAssembly -DesiredVersion $NugetPackageVersion)
+        {
+            Write-Log "Found $AssemblyName in module directory ($PSScriptRoot)." -Level Verbose
+            return $moduleAssembly
+        }
+        else
+        {
+            Write-Log "Found $AssemblyName in module directory ($PSScriptRoot), but its version number didn't match required [$NugetPackageVersion]." -Level Verbose
```
danbelcher-MSFT: This seems like something we should always want the user to see, no?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -732,17 +807,27 @@ function Set-SubmissionPackage
...
+                    $bytes = [System.IO.File]::ReadAllBytes($AzureStorageDataMovementDll)
+                    [System.Reflection.Assembly]::Load($bytes) | Out-Null
+
                     $uri = New-Object -TypeName System.Uri -ArgumentList $UploadUrl
                     $cloudBlockBlob = New-Object -TypeName Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob -ArgumentList $uri
-                    $cloudBlockBlob.UploadFromFile($PackagePath, [System.IO.FileMode]::Open)
+
+                    # We will run this async command synchronously within the console.
+                    $task = [Microsoft.WindowsAzure.Storage.DataMovement.TransferManager]::UploadAsync($PackagePath, $cloudBlockBlob, $null, $null)
+                    $task.GetAwaiter().GetResult() | Out-Null
```
danbelcher-MSFT: Can't we avoid this repeated code by putting it in a function and either calling that function or starting a job that calls that function?  The job will have access to functions defined in this script, right?

StoreBroker/NugetTools.ps1
HowardWolosky: good suggestion. done.

StoreBroker/NugetTools.ps1
HowardWolosky: You're right that no-one is using it.  I had added it because I was considering using it in a different case.  I'll just remove it for now and add that logic back in if/when I do need it.

StoreBroker/NugetTools.ps1
HowardWolosky: Good catch, fixed.

StoreBroker/NugetTools.ps1
HowardWolosky: _Won't fix_: Nah.  Because they might see it up to three times (alt path, assembly path, temp path).  They'll still see the message about how to update their copy so that they can avoid the download in the future, so I don't see a reason to add that to normal output.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Jobs _don't_ have access to functions in the module.  They don't even have access to local vars defined just before the job was created.  That's why I have to pass in the local vars that have the path to the assemblies...The job executes in a completely fresh PowerShell context.  So no...I won't be pulling this out/refactoring.  The code would actually become even more convoluted if I tried.

PR #54, opened 2017-06-28
Update ProductDescription.xml

egcorral:
Updated the MaxLength value according to recent Store changed requirements (6/14/2017 - Store Requirement Change | Search Terms). It is an important change that impacts Localization.

Old	New
Up to 7 Search terms	Up to 7 Search terms
45 Character limit for each search term	30 Character limit for each search term
No more than 21 separate words across search terms	No more than 21 separate words across search terms

### DIFF
--- a/PDP/ProductDescription.xml
+++ b/PDP/ProductDescription.xml
@@ -12,11 +12,11 @@
 
   <Keywords>
     <!-- Valid length: 45 character limit, up to 7 -->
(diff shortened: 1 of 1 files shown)


danbelcher-MSFT:
Hi Erica,

I appreciate you being an active participant in StoreBroker.  Thank you!  Unfortunately, it looks like there is already a pull request made to address this issue, see #51 .  I'm not sure why that change hasn't been checked in yet; I will talk with Howard so we can close on the issue.  For your change, I am going to reject the pull request in favor of the other.  Our apologies for the inconvenience of duplicated effort here.

Thanks,
-Daniel Belcher

PR #57, opened 2017-07-11
Jowis/modify halt exception messages

wisniew41:
Removing "Halt Execution" exception messages and replacing them with the output message or throwing the last exception message.
Also added the MS-correlation Id to the output when applicable.

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -523,7 +523,7 @@ function Test-Xml
         $msg = $msg -join [Environment]::NewLine
 
         Write-Log $msg -Level Error
-        throw "Halt Execution"
+        throw $msg
     }
 }
 
@@ -697,8 +697,9 @@ function Convert-ListingToObject
 
                                     if ($null -eq $image)
                                     {
-                                        Write-Log "Could not find image '$($imageFileName)' in any subdirectory of '$imageContainerPath'." -Level Error
-                                        throw "Halt Execution"
+                                        $output = "Could not find image '$($imageFileName)' in any subdirectory of '$imageContainerPath'."
+                                        Write-Log $output -Level Error
+                                        throw $output
                                     }
                         
                                     $destinationInPackage = Join-Path $packageImagePath $image.Name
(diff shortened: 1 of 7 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -2009,7 +2015,7 @@ function Get-SubmissionRequestBody
                 $out += "Check the values of '$script:s_PDPRootPath' and '$script:s_Release' and try again."
 
                 Write-Log ($out -join [Environment]::NewLine) -Level Error
```
danbelcher-MSFT: Write this as you wrote the others.  Save the ($out -join Newline) expression.  Pass that variable to write-log and then throw that variable.

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1929,13 +1930,13 @@ function Invoke-SBRestMethod
         {
             Write-Log $_.Exception.Message -Level Error
             Set-TelemetryException -Exception $_.Exception -ErrorBucket $errorBucket -Properties $localTelemetryProperties
-            throw;
+            throw $_.Exception.Message + '. ' + $script:headerMSCorrelationId + ':' + $correlationId;
```
danbelcher-MSFT: I would reverse the order of this concatenation.  The length of the exception message is variable and potentially quite large.  Start with the correlation Id because that length is known and relatively small.  Follow that with the exception message.

Additionally, you're concatenating these with a ". " separator.  Please use [Environment]::Newline instead so they appear on separate lines.  Using ". " makes it easier to miss the correlation Id (again, doubly so if the Id is at the end of the exception message)

StoreBroker/PackageTool.ps1
wisniew41: Added this to the $newLineOutput variable where I am doing this

StoreBroker/StoreIngestionApi.psm1
wisniew41: I swapped these to have the correlation Id appear first and used the [Environment]::Newline as the separator instead for this

StoreBroker/PackageTool.ps1
```diff
@@ -2274,7 +2284,7 @@ function Resolve-PackageParameters
             if ([System.String]::IsNullOrWhiteSpace($configVal))
             {
                 Write-Log ($out -f $script:s_OutName) -Level Error
-                throw "Halt Execution"
+                throw ($out -f $script:s_OutName)
```
danbelcher-MSFT: missed one

wisniew41: shoot, it has now been changed

PR #62, opened 2017-09-28
Process ID (PID) of StoreBroker PowerShell session now captured in log

HowardWolosky:
Some users have scenarios where they are running concurrent instances of StoreBroker,
and for reasons specific to their situation, have chosen to have the logs always go
to the same log file (instead of using `$global:SBLogPath` to set a different
log file for each session.

In these scenarios, it can be challenging to read the logs and understand what exactly
was happening since there are entries from multiple commands intermixed.  This
change introduces adding the `$pid` of the StoreBroker PowerShell session to every
log entry so that on post-mortem analysis, it would be easy to filter to just the
logs for the specific `$pid`.

For more info on `$pid`, [see here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1&viewFallbackFrom=powershell-Microsoft.PowerShell.Core#pid)

### DIFF
--- a/Documentation/USAGE.md
+++ b/Documentation/USAGE.md
@@ -72,6 +72,18 @@ when the module is loaded.
     When `$true`, times are logged using UTC (and those timestamps will end with a Z per the
     [W3C standard](http://www.w3.org/TR/NOTE-datetime))
 
+ **`$global:SBShouldLogPid`** [bool] Defaults to `$false`. If `$true`, the
+ Process ID (`$global:PID`) of the current PowerShell process will be added
+ to every log entry.  This can be helpful if you have situations where
+ multiple instances of StoreBroker run concurrently and you want to
+ more easily isolate the log entries for one process.  An alternative
+ solution would be to use `$global:SBLogPath` to specify a different
+ log file for each StoreBroker process. An easy way to view the filtered
+ entries for a session is (replacing `PID` with the PID that you are interested in):
+
+    Get-Content -Path $global:SBLogPath -Encoding UTF8 | Where { $_ -like '*[[]PID[]]*' }
+
+
 > **PowerShell Tip**
 >
 > If you wish to always use a different value for these, set the new values in your PowerShell
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -52,6 +52,11 @@ function Initialize-HelpersGlobalVariables
         }
     }
 
+    if (!(Get-Variable -Name SBShouldLogPid -Scope Global -ValueOnly -ErrorAction SilentlyContinue))
+    {
+        $global:SBShouldLogPid = $false
+    }
+
     if (!(Get-Variable -Name SBNotifyDefaultDomain -Scope Global -ValueOnly -ErrorAction SilentlyContinue))
     {
         $global:SBNotifyDefaultDomain = $null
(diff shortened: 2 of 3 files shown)


Documentation/USAGE.md
```diff
@@ -72,6 +72,18 @@ when the module is loaded.
     When `$true`, times are logged using UTC (and those timestamps will end with a Z per the
     [W3C standard](http://www.w3.org/TR/NOTE-datetime))
 
+ **`$global:SBLogPid`** [bool] Defaults to `$false`. If `$true`, the
```
danbelcher-MSFT: The name by itself sounds like a numerical value, or a variable for a Pid to use for logging.  Maybe 'SBShouldLogPid'?

StoreBroker/Helpers.ps1
```diff
@@ -483,18 +488,42 @@ function Write-Log
...
-            $Level.ToUpper(),
-            $Message
+        if ($global:SBLogPid)
+        {
+            $MaxPidDigits = 10 # This is an estimate (see https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows)
+            $pidPadding = $MaxPidDigits - "$pid".length 
+            $logFileMessage = '{0}{1} : [{2}]{3}: {4} : {5} : {6}' -f
+                (" " * $Indent),
+                $dateString,
+                $pid,
```
danbelcher-MSFT: $pid is the same syntax we use for local variables.  Can we go with `$global:PID`?

StoreBroker/Helpers.ps1
```diff
@@ -483,18 +488,42 @@ function Write-Log
...
-        $logFileMessage = '{0}{1} : {2} : {3} : {4}' -f
-            (" " * $Indent),
-            $dateString,
-            $env:username,
-            $Level.ToUpper(),
-            $Message
+        if ($global:SBLogPid)
+        {
+            $MaxPidDigits = 10 # This is an estimate (see https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows)
+            $pidPadding = $MaxPidDigits - "$pid".length
```
danbelcher-MSFT: There should be format specifiers for adding appropriate padding.  Would save you a variable and some logic.

danbelcher-MSFT: PS C:\> $MaxPidDigits = 10
PS C:\> "'{0,$MaxPidDigits}'" -f 12341
'     12341'

danbelcher-MSFT: Well, GitHub removed the extra spaces, but you get the idea :)

StoreBroker/Helpers.ps1
```diff
@@ -483,18 +488,42 @@ function Write-Log
...
 
-        $logFileMessage = '{0}{1} : {2} : {3} : {4}' -f
-            (" " * $Indent),
-            $dateString,
-            $env:username,
-            $Level.ToUpper(),
-            $Message
+        if ($global:SBLogPid)
+        {
+            $MaxPidDigits = 10 # This is an estimate (see https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows)
```
danbelcher-MSFT: Shouldn't this have $maxPidDigits syntax?

Since we're estimating, it seems safer to go with:
$maxPidDigits = [Math]::Max(10, "$global:PID".Length)

Just in case.

Documentation/USAGE.md
HowardWolosky: Sounds reasonable.

StoreBroker/Helpers.ps1
HowardWolosky: indeed

StoreBroker/Helpers.ps1
HowardWolosky: I was considering this a constant, hence SentenceCase as opposed to camelCase (for a varaible).  Your suggestion on using `Max` for this scenario sounds good though.

StoreBroker/Helpers.ps1
HowardWolosky: While interesting, that doesn't get me the formatting that I want.  I want the padding on the end, not the beginning.

Issue #64, opened 2017-10-02
Convert relative OutPath to absolute path in Ensure-PdpFilePath


martinsuchan:
In **ConvertFrom-ExistingSubmission.ps1** if user enters **relative -OutPath** and launches the script from non-default folder, the OutPath folder is created in a wrong place and the **$xml.Save($filePath) throws an exception** because it can't find the target nested folder.
```
.\ConvertFrom-ExistingSubmission.ps1 -AppId "ABCDEFGH" -Release "Master" -PdpFileName "PDP.xml" -OutPath "PDPOut"
```
This creates the OutPath folder in Extensions directory, but $xml.Save($filePath) expects it in StoreBroker root directory.

Expected solution: Ensure-PdpFilePath should return absolute path where the OutPath folder was created.

HowardWolosky:
Thanks for the report.
Looks like we should be using `Resolve-UnverifiedPath` here since it's calling out to a non-PowerShell command (`$xml.Save()`).  That command was introduced in [this commit](https://github.com/Microsoft/StoreBroker/commit/2a86150d134d3f0a1552b00b199dd252a6092a04) because package upload/download also use non-native command which means that they don't handle relative paths correctly.

@danbelcher-MSFT - Can you update those scripts to leverage `Resolve-UnverifiedPath`, and then make sure that there aren't any other code-paths in StoreBroker that take in a path to a non-native PowerShell method that haven't been resolved first?

Thanks.

danbelcher-MSFT:
Closing this issue as resolved based on the linked commit.  Please let us know if you are still experiencing this issue.  Thanks!

PR #67, opened 2017-10-02
Support relative paths that don't begin with '.'

danbelcher-MSFT:
Support relative paths that don't begin with '.' to denote the current working directory.

For example, an -OutPath of "SomeFolder" should represent a sub-folder of the current working directory. While PowerShell can handle path fragments of this type, interacting with .NET methods, such as XmlDocument.Save() can cause the file to be saved to an unintended directory.

The fix is to use Resolve-UnverifiedPath in-order to resolve path fragments like this to a full path.

Additonally, adds a new function Ensure-Directory, to ensure a directory is created/exists.

#64

### DIFF
--- a/Extensions/ConvertFrom-ExistingIapSubmission.ps1
+++ b/Extensions/ConvertFrom-ExistingIapSubmission.ps1
@@ -601,5 +601,10 @@ function Main
 }
 
 
+
+
+# Script body
+$OutPath = Resolve-UnverifiedPath -Path $OutPath
+
 # function Main invocation
 Main
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -839,3 +839,44 @@ function Resolve-UnverifiedPath
         return $resolvedPath.ProviderPath
     }
 }
+
+function Ensure-Directory
+{
+<#
+    .SYNOPSIS
+        A utility function for ensuring a given directory exists.
+
+    .DESCRIPTION
+        A utility function for ensuring a given directory exists.
+
+        If the directory does not already exist, it will be created.
+
+    .PARAMETER Path
+        A full or relative path to the directory that should exist when the function exits.
+
+    .NOTES
+        Uses the Resolve-UnverifiedPath function to resolve relative paths.
+#>
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory)]
+        [string] $Path
+    )
+
+    try
+    {
+        $Path = Resolve-UnverifiedPath -Path $Path
+
+        if (-not (Test-Path -PathType Container -Path $Path))
+        {
+            Write-Log "Creating directory: [$Path]" -Level Verbose
+            New-Item -ItemType Directory -Path $Path | Out-Null
+        }
+    }
+    catch
+    {
+        Write-Log "Could not ensure directory: [$Path]" -Level Error
+
+        throw 
+    }
+}
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -516,6 +516,8 @@ function Write-SubmissionRequestBody
 
     if ($PSCmdlet.ShouldProcess($OutFilePath, "Output to File"))
     {
+        Ensure-Directory -Path (Split-Path -Parent -Path $OutFilePath)
+
         Write-Log "Writing submission request JSON file: [$OutFilePath]." -Level Verbose
 
         $JsonObject |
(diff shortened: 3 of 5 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -2907,7 +2909,6 @@ function New-SubmissionPackage
         })]
         [string[]] $AppxPath,
 
-        [ValidateScript({ if (Test-Path -PathType Container $_) { $true } else { throw "$_ cannot be found." } })]
```
danbelcher-MSFT: Removing the existence check for OutPath here and below.  Non-existent paths are still valid and can be created during the output step.

HowardWolosky:
You should clarify in your commit message that the issue really only can be seen when using a .NET method (like `[xml]::Save()`) with a relative path.  Relative paths work just fine when you stay completely within PowerShell commands...

StoreBroker/Helpers.ps1
```diff
@@ -838,3 +838,48 @@ function Resolve-UnverifiedPath
...
+
+    try
+    {
+        $resolvedPath = Resolve-UnverifiedPath -Path $Path
+        if ($null -ne $resolvedPath)
+        {
+            $Path = $resolvedPath
+        }
+
+        if (-not (Test-Path -PathType Container -Path $Path ))
```
HowardWolosky: nit: extra space before the closing parens

StoreBroker/Helpers.ps1
```diff
@@ -838,3 +838,48 @@ function Resolve-UnverifiedPath
...
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory)]
+        [string] $Path
+    )
+
+    try
+    {
+        $resolvedPath = Resolve-UnverifiedPath -Path $Path
+        if ($null -ne $resolvedPath)
```
HowardWolosky: What's the scenario when this will be `$null`? (besides of course the input value being `$null`)

Extensions/ConvertFrom-ExistingIapSubmission.ps1
```diff
@@ -539,6 +539,62 @@ function Show-ImageFileNames
     }
 }
 
+function Resolve-UnverifiedPath
```
HowardWolosky: I know that we wanted these scripts to work independent of the module, but now I'm not 100% sure.  Having to duplicate this logic in here now across three different files seems like a maintenance nightmare.  What's the drawback of dot-sourcing Helpers.ps1 here?

StoreBroker/Helpers.ps1
```diff
@@ -829,7 +829,7 @@ function Resolve-UnverifiedPath
 
     $resolvedPath = Resolve-Path -Path $Path -ErrorVariable resolvePathError -ErrorAction SilentlyContinue
 
-    if ($null -eq $resolvedPath)
+    if (($null -eq $resolvedPath) -and ($resolvePathError.Count -gt 0))
```
HowardWolosky: Can you explain the scenario for this change?

Extensions/ConvertFrom-ExistingIapSubmission.ps1
danbelcher-MSFT: In fact, I forgot we are already dot-sourcing Helpers.ps1 at the top of the file.  I'll remove the redefinition of the function.

StoreBroker/Helpers.ps1
danbelcher-MSFT: This came about after testing with -ErrorAction Ignore.  In that case, $resolvePathError was still $null and $resolvedPath was also $null.  An exception was thrown trying to index into $resolvePathError.  After changing back to -ErrorAction SilentlyContinue, I left the additional guard in just to be safe.

I'm fine removing if you don't think we need it.

StoreBroker/Helpers.ps1
danbelcher-MSFT: This came about from testing with -ErrorAction Ignore, as explained above.  I can remove the check.

StoreBroker/Helpers.ps1
HowardWolosky: It's not going to help anything in this case...you'll just force it to go in to the else, and then it'll try accessing a property off of `$null` and get back an empty string, which will end up causing issues later on when the caller tries to use that "resolved" path.  Better to just let it crash here if it really doesn't have any errors that we can index into.

StoreBroker/Helpers.ps1
HowardWolosky: We really shouldn't be in a situation where `Resolve-UnverifiedPath` is returning `$null`.  If it did, I'm not sure that just silently falling back to using the original `$Path` is the best idea....it just means that the caller is not guaranteed that they'll be getting their file written to the expected location...especially since the .zip file is going to run into trouble since it also depends on a valid value coming back from `Resolve-UnverifiedPath`

PR #70, opened 2017-10-13
Fix Additional Metadata for App Submission packages

HowardWolosky:
In v1.6.0, we added [additional metadata](https://github.com/Microsoft/StoreBroker/commit/391954b962756b1f222d2e4e45aa6d203b805db4)
to the generated JSON file for application submission packages.

Unfortunately, we operated under the incorrect assumption that there could
only ever be a single Min OS version specified for a package.  In fact, a
single package can target multiple device families, and each device family
can target a different minimum OS version.  The only requirement around
Min OS Versions for packages is that if there is an appxbundle, all appx
files within that bundle must target the same minimum os version for a given
device family.

Given this change in our understanding of how things work, we're modifying
the additional metadata that we were generating.  `targetDeviceFamiliesEx`
will now contain the same version that `targetDeviceFamilies` contains
(both the device family name and the minimum os version), but they will
be split into named properties.  Additionally, the `minOsVersion` property
has been removed since the data it has is less relevant when it's not paired
with the specific deviceFamily that it's related to.

Finally, we are now adding a new `sbschema` property (which is simply an
integer) which we'll be increasing from now on whenever we make changes
to non-standard properties in our output JSON in the event that StoreBroker
(or other applications leveraging the StoreBroker payload) are depending on
those values and need to know how to behave if they're not there.

USAGE.md has been updated to now track (and explain) the
[schema version changes](https://github.com/Microsoft/StoreBroker/blob/master/Documentation/USAGE.md#schema-versions)
(for botn App submissions and for IAP submissions).

### DIFF
--- a/Documentation/USAGE.md
+++ b/Documentation/USAGE.md
@@ -24,6 +24,7 @@
     *   [IAP Overview](#iap-overview)
     *   [Creating Your IAP Payload](#creating-your-iap-payload)
     *   [IAP Commands](#iap-commands)
+*   [Schema Versions](#schema-versions)
 *   [Using INT vs PROD](#using-int-vs-prod)
 *   [Telemetry](#telemetry)
 *   [FAQ](#faq)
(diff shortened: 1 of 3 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1453,14 +1459,15 @@ function Read-AppxMetadata
             Sort-Object -Unique
 
         $metadata.targetDeviceFamiliesEx = @()
-        $metadata.targetDeviceFamiliesEx += $manifest.Package.Dependencies.TargetDeviceFamily.Name |
-            Where-Object { $null -ne $_ } |
-            Sort-Object -Unique
+        $metadata.targetDeviceFamiliesEx += $manifest.Package.Dependencies.TargetDeviceFamily |
```
danbelcher-MSFT: I generally just wrap the pipeline with @() to keep this as one line. Up to you.

StoreBroker/PackageTool.ps1
```diff
@@ -1677,21 +1684,20 @@ function Read-AppxBundleMetadata
...
             $metadata.innerPackages.$($appxMetadata.architecture) = @{
-                version              = $appxMetadata.version;
-                targetDeviceFamilies = $appxMetadata.targetDeviceFamiliesEx;
-                languages            = $appxMetadata.languages;
-                capabilities         = $appxMetadata.capabilities;
-                minOSVersion         = $appxMetadata.minOSVersion;
-                targetPlatform       = $appxMetadata.targetPlatform;
+                version                = $appxMetadata.version;
+                targetDeviceFamiliesEx = $appxMetadata.targetDeviceFamiliesEx
+                targetDeviceFamilies   = $appxMetadata.targetDeviceFamiliesEx | ForEach-Object { "$($_.name) min version $($_.minOSVersion)" }
```
danbelcher-MSFT: The "{0} min version {1}" string is duplicated here and above.  Worth making a constant?  Or worth making a function to return the formatted string?

Documentation/USAGE.md
```diff
@@ -879,6 +880,104 @@ To monitor an IAP submission:
 Follow the steps in [monitoring a submission](#monitoring-a-submission), and be sure to include
 **`-IapId`** in the function parameters _instead of_ **`-AppId`**.
 
+## Schema Versions
+
+StoreBroker's packaging commands (`New-SubmissionPackage` and `New-InAppProductSubmissionPackage`)
+add proprties to the generated .json file that are not part of the official Submission API JSON
```
danbelcher-MSFT: typo: "properties"

StoreBroker/PackageTool.ps1
```diff
@@ -33,10 +33,17 @@ $script:applicationMetadataProperties = @(
...
     "innerPackages"
 )
 
+# The current version of the StoreBroker schema that PackageTool is authoring for app and IAP submissions.
+# The StoreBroker schema may include metadata that isn't a core part of the official Submission API
+# JSON schema (like the appId or iapId, package metadata, etc...). These values should be updated any time
+# we alter what additional metadata is added to the schema for that submission type.
+$script:appSchemaVersion = 2;
+$script:iapSchemaVersion = 2;
+$script:schemaPropertyName = 'sbschema'
```
danbelcher-MSFT: camel-case?  'sbSchema'

Documentation/USAGE.md
```diff
@@ -879,6 +880,104 @@ To monitor an IAP submission:
...
+it necessary.
+
+To that end, starting with version `1.11.1` of StoreBroker, we will always add a `sbschema` property
+to the generated json file, with a simple integer value that will be incremented whenever we make any
+change to those additional properties.  We'll document what those differences are below so that you'll
+be able to make corresponding changes to your own code if you depend on these properties.
+
+### App Submission Packages
+
+#### 1
```
danbelcher-MSFT: I'd prefer 'Version 1' explicitly. Having just '1' looks odd in the rendered markdown.

Documentation/USAGE.md
```diff
@@ -879,6 +880,104 @@ To monitor an IAP submission:
...
+
+> We never used the value `1` for a published schema.  Any App JSON package file that doesn't have an
```
danbelcher-MSFT: I think this is okay as regular text and not as a note. Ver 2 has regular text here. On GitHub, the note uses gray font color and is actually harder to read than regular text.

StoreBroker/PackageTool.ps1
```diff
@@ -33,10 +33,17 @@ $script:applicationMetadataProperties = @(
...
     "targetDeviceFamiliesEx",
-    "minOSVersion",
     "innerPackages"
 )
 
+# The current version of the StoreBroker schema that PackageTool is authoring for app and IAP submissions.
+# The StoreBroker schema may include metadata that isn't a core part of the official Submission API
+# JSON schema (like the appId or iapId, package metadata, etc...). These values should be updated any time
+# we alter what additional metadata is added to the schema for that submission type.
+$script:appSchemaVersion = 2;
```
danbelcher-MSFT: nit: don't need the semi-colons

danbelcher-MSFT: Maybe we should define these in Helpers.ps1? Or other shared file?  We don't have a need to reference these elsewhere right now but they also aren't necessarily only useful for PackageTool

Documentation/USAGE.md
HowardWolosky: fixed.

Documentation/USAGE.md
HowardWolosky: The reason that I had done that was so that it mapped to the literal value that the property would have.  I agree that it looked weird though.  I've added the word `Version` before the `1` and `2` and I **bolded** the **`integer value`** note above that area indicating what the actual value of the `sbschema` could be.

Documentation/USAGE.md
HowardWolosky: done

StoreBroker/PackageTool.ps1
HowardWolosky: Not sure how that slipped in there.

HowardWolosky: It's a good thought.  These _are_ accessible to any of the other scripts.  I _don't_ want to put them in Helpers.  My general hope for Helpers.ps1 is that it eventually goes away and those generic methods come from some other shared project.

I'd much rather that these "constants" be defined and live close to the code that is using them at its core (in this case, PackageTool).

StoreBroker/PackageTool.ps1
HowardWolosky: Sure. Done.

StoreBroker/PackageTool.ps1
HowardWolosky: Fine with leaving this as-is for now, given that this current pattern is used elsewhere in even this method, so I'd prefer consistency in this case (and I'd rather not make unrelated changes within the context of this PR).

StoreBroker/PackageTool.ps1
HowardWolosky: Yep. Good idea.

PR #73, opened 2017-10-26
Capture activityId to log with 500 / Internal Server Error failures

### LINKED ISSUE #72: Add activityId to log when API call fails with 500 / Internal Server Error
Issues like we've seen with #71 would apparently be easier to diagnose on the API side if we were able to provide the `activityId` property associated with the API call.  That property is only ever seen when there is a 500 / Internal Server error, and the response would look something like this:

`{"statusCode":500,"message":"Internal server error","activityId":"abcdef01-2345-6789-abcd-ef0123456789"}`

HowardWolosky:
According to the API team, in the cases where they return back a
500 / Internal Server Error, they will return back JSON content
in the body that contains an `activityId` property, which will be
a GUID.  This property will apparantly help the Submission API team
diagnose these server errors.  Therefore, we now try to log it.

A `WebException` contains an `HttpWebResponse` object which requires
stream access in order to read any embedded content.  So, I've added
a new helper method, `Get-HttpWebResponseContent` which we'll call
when accessing the `WebException` in order to get to the raw content.

If we are able to get content, deserialize it, _and_ there's an
`activityId`, we'll just log that `activityId`.  Otherwise, we'll
log the entire raw content (if it exists).

I'm also reverting a small part of change 6c8bed398, which attempted
to throw a custom error message containing the `MS-CorrelationId` in
a scenario where we would never actually have it.  In this case, I'm
simply reverting that code to just throw the original exception.

This also fixes any dangling PSScriptAnalyzer issues that _can_ be fixed,
and explicitly notes in the code the ones that can't be fixed
due to (referenced) PSScriptAnalyzer bugs.

Resolves #72: Add activityId to log when API call fails with 500 / Internal Server Error

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -27,6 +27,9 @@ function Initialize-HelpersGlobalVariables
     [CmdletBinding()]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidGlobalVars", "", Justification="We use global variables sparingly and intentionally for module configuration, and employ a consistent naming convention.")]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseSingularNouns", "", Justification="We are initializing multiple variables.")]
+
+    # Note, this doesn't currently work due to https://github.com/PowerShell/PSScriptAnalyzer/issues/698
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseDeclaredVarsMoreThanAssignment", "", Justification="These are global variables and so are used elsewhere.")]
     param()
 
     # We only set their values if they don't already have values defined.
@@ -858,6 +861,7 @@ function Ensure-Directory
         Uses the Resolve-UnverifiedPath function to resolve relative paths.
 #>
     [CmdletBinding()]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseApprovedVerbs", "", Justification = "Unable to find a standard verb that satisfies describing the purpose of this internal helper method.")]
     param(
         [Parameter(Mandatory)]
         [string] $Path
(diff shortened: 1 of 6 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -880,3 +880,57 @@ function Ensure-Directory
...
+        you get a WebException which contains a simpler WebResponse, which
+        requires a bit more effort in order to acccess the raw response content.
+
+    .PARAMETER WebResponse
+        An HttpWebResponse object, typically the Response property on a WebException.
+
+    .OUTPUTS
+        System.String - The raw content that was included in a WebResponse; $null otherwise.
+#>
+    [CmdletBinding()]
```
danbelcher-MSFT: Did you want to use [OutputType()] attribute?

StoreBroker/Helpers.ps1
```diff
@@ -880,3 +880,57 @@ function Ensure-Directory
...
+    param(
+        [Parameter(Mandatory)]
+        [System.Net.HttpWebResponse] $WebResponse
+    )
+
+    $streamReader = $null
```
danbelcher-MSFT: These null initializations aren't actually needed but I think you included them for completeness?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1792,7 +1792,12 @@ function Invoke-SBRestMethod
             if ($PSCmdlet.ShouldProcess($jobName, "Start-Job"))
             {
                 [scriptblock]$scriptBlock = {
-                    param($Url, $method, $Headers, $Body, $HeaderName, $TimeoutSec)
+                    param($Url, $method, $Headers, $Body, $HeaderName, $TimeoutSec, $ScriptRootPath)
```
danbelcher-MSFT: Opportunity to switch to $using:var?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1838,7 +1844,8 @@ function Invoke-SBRestMethod
                     }
                 }
 
-                $null = Start-Job -Name $jobName -ScriptBlock $scriptBlock -Arg @($url, $Method, $headers, $Body, $script:headerMSCorrelationId, $global:SBWebRequestTimeoutSec)
+                $scriptRootPath = Split-Path $MyInvocation.PSCommandPath -Parent
```
danbelcher-MSFT: If my understanding is right, .PSCommandPath is the path of the script that _called_ the executing script, whereas .MyCommand.Path is for the path of the script being executed.  Is this still correct?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1898,6 +1905,7 @@ function Invoke-SBRestMethod
         $statusDescription = $null
         $correlationId = $null
         $innerMessage = $null
+        $rawContent = $null
```
danbelcher-MSFT: Why are we initializing variables with $null?  Are the variables pre-existing?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1973,6 +1983,37 @@ function Invoke-SBRestMethod
...
+        # If it did, we want to extract the "activityId" property from it for
+        # logging purposes in order to assist the Submission API team with 
+        # post-mortem debugging.
+        if (-not [String]::IsNullOrWhiteSpace($rawContent))
+        {
+            try
+            {
+                $rawContentJson = $rawContent | ConvertFrom-Json
+                $activityId = $rawContentJson.activityId
+                if ($null -ne $activityId)
```
danbelcher-MSFT: I'd prefer -not [String]::IsNullOrEmpty()  in-case the API returns an empty activityId

HowardWolosky: Indeed.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: It's for [PSScriptAnalyzer](https://github.com/PowerShell/PSScriptAnalyzer).  It complains if it thinks you're using an unassigned variable, and it has some issues when it comes to scope.  Those variables are getting their true assignments within if blocks but then used at one scope higher.  So, by assigning a value to them (in this case, `$null`), at the same scope level as their usage, we avoid the spurious error.  Otherwise, there'd be so much noise in the results that we'd simply stop using it altogether.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Did some more digging....I can just use `$PSScriptRoot`.  That gives me the path for the current executing script.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Possibly, but I'm not inclined to do that as part of this change, especially now that I switched to just passing in `$PSScriptRoot`

StoreBroker/Helpers.ps1
HowardWolosky: As mentioned elsewhere, this is mostly to satisfy [PSScriptAnalyzer](https://github.com/PowerShell/PSScriptAnalyzer) and avoid spurious errors.

StoreBroker/Helpers.ps1
HowardWolosky: Sure. Done.

PR #76, opened 2017-10-31
Fix for trying to get the RawStream for errors.

HowardWolosky:
36384212b introduced a change to try to get additional error
content from the response stream when we received an error from
the server.

Unfortunately, there was a typo in the parameter name we were
attempting to access the content stream from, resulting in
a new exception being thrown: "You cannot call a method on a
null-valued expression".

This fixes the typo, and also makes the attempt to get that value
best-effort by doing a try/catch around the call, simply so that
we never lose reporting the true error from the API.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -922,8 +922,12 @@ function Get-HttpWebResponseContent
 
         if ($WebResponse.ContentLength -gt 0)
         {
-            $stream = $Response.GetResponseStream()
-            $encoding = [System.Text.Encoding]::GetEncoding($WebResponse.ContentEncoding)
+            $stream = $WebResponse.GetResponseStream()
+            $encoding = [System.Text.Encoding]::UTF8
+            if (-not [String]::IsNullOrWhiteSpace($WebResponse.ContentEncoding))
+            {
+                $encoding = [System.Text.Encoding]::GetEncoding($WebResponse.ContentEncoding)
+            }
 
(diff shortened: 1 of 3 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -923,7 +923,11 @@ function Get-HttpWebResponseContent
         if ($WebResponse.ContentLength -gt 0)
         {
             $stream = $WebResponse.GetResponseStream()
-            $encoding = [System.Text.Encoding]::GetEncoding($WebResponse.ContentEncoding)
+            $encoding = [System.Text.Encoding]::UTF8
```
danbelcher-MSFT: single assignment?  $encoding = if () {} else {}

HowardWolosky: Line would be too long.  Was more readable this way.

danbelcher-MSFT: Okay, I won't nit-pick if you want to keep it, but I think the length issue is easily solved, either:
```
if ()
{
    $encoding = ...
}
else
{
    $encoding = ...
}
```
or
```
$encoding = if () {}
            else {}
```

HowardWolosky: Your first suggestion falls into the issue that PSScriptAnalyzer would incorrectly flag us as using an uninitialized variable.  So then we'd initialize `$encoding` before the if check, and it would start looking like what I already have.  The second suggestion goes against our standard coding conventions of having brackets on individual lines (we do intentionally violate that one in some specific situations, but this doesn't seem like a worthwhile scenario.

This method initializes our variable to a value that we like, and only changes it if we know we can get a better one from our input.

PR #77, opened 2017-11-10
Fix how we reference Special folders (Desktop, Documents)

HowardWolosky:
We previously referenced the Desktop and Documents folders like so:
`$home\Documents` and `$home\desktop`, but those folders can be
redirected to a different drive (and folder name) without changing
the user's profile directory (e.g. `$home`).

Given this, we will use the special method
[[System.Environment]::GetFolderPath()](https://msdn.microsoft.com/en-us/library/14tx8hby%28v=vs.110%29.aspx)
which can give us the correct path to those ["special folders"](https://msdn.microsoft.com/en-us/library/system.environment.specialfolder(v=vs.110).aspx)
(named `MyDocuments` and `Desktop`).

This required minor changes to the code and the documentation.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -41,14 +41,12 @@ function Initialize-HelpersGlobalVariables
         $global:SBLoggingEnabled = $true
     }
 
-    # $Home relies on existence of $env:HOMEDRIVE and $env:HOMEPATH which are only 
-    # set when a user logged in interactively, which may not be the case for some build machines.
-    # $env:USERPROFILE is the equivalent of $Home, and should always be available.
     if (!(Get-Variable -Name SBLogPath -Scope Global -ValueOnly -ErrorAction Ignore))
     {
-        if (-not [System.String]::IsNullOrEmpty($env:USERPROFILE))
+        $documentsFolder = [System.Environment]::GetFolderPath('MyDocuments')
+        if (-not [System.String]::IsNullOrEmpty($documentsFolder))
         {
-            $global:SBLogPath = $(Join-Path $env:USERPROFILE "Documents\StoreBroker.log")
+            $global:SBLogPath = Join-Path -Path $documentsFolder -ChildPath 'StoreBroker.log'
         }
         else
         {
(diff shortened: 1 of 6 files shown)


StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -1077,13 +1077,15 @@ function Update-ApplicationSubmission
...
+        $configPath = Join-Path -Path ([System.Environment]::GetFolderPath('Desktop')) -ChildPath 'newconfig.json'
+
         $output = @()
         $output += "The config file used to generate this submission did not have an AppId defined in it."
         $output += "The AppId entry in the config helps ensure that payloads are not submitted to the wrong application."
         $output += "Please update your app's StoreBroker config file by adding an ""appId"" property with"
         $output += "your app's AppId to the ""appSubmission"" section.  If you're unclear on what change"
         $output += "needs to be done, you can re-generate your config file using"
-        $output += "   ""New-StoreBrokerConfigFile -AppId $AppId"" -Path ""`$home\desktop\newconfig.json"""
+        $output += "   ""New-StoreBrokerConfigFile -AppId $AppId -Path ""$configPath"""
```
danbelcher-MSFT: nit: I prefer the back-tick + double quote syntax over 2x double quote because it takes away the need to mentally parse the string.

StoreBroker/Helpers.ps1
```diff
@@ -41,14 +41,12 @@ function Initialize-HelpersGlobalVariables
...
-    # set when a user logged in interactively, which may not be the case for some build machines.
-    # $env:USERPROFILE is the equivalent of $Home, and should always be available.
     if (!(Get-Variable -Name SBLogPath -Scope Global -ValueOnly -ErrorAction Ignore))
     {
-        if (-not [System.String]::IsNullOrEmpty($env:USERPROFILE))
+        $documentsFolder = [System.Environment]::GetFolderPath('MyDocuments')
+        if (-not [System.String]::IsNullOrEmpty($documentsFolder))
         {
-            $global:SBLogPath = $(Join-Path $env:USERPROFILE "Documents\StoreBroker.log")
+            $global:SBLogPath = (Join-Path -Path $documentsFolder -ChildPath 'StoreBroker.log')
```
danbelcher-MSFT: In this CR, when we're assigning the result of Join-Path to a variable, we're inconsistent about whether the expression is wrapped in parens or not.

HowardWolosky: Ah, good catch.  There's no need for that outer parenthesis...I'll remove it.

StoreBroker/StoreIngestionApplicationApi.ps1
HowardWolosky: Fair feedback.  Will fix.

PR #79, opened 2017-11-30
Provide helpful error message if users try to manipulate packages without specifying any

HowardWolosky:
If a user tries to use `-AddPackages` or `-ReplacePackages` with a StoreBroker payload
that doesn't contain any, they end up with a literal `null` entry in the `applicationPackages`
list due to how we manipulate the submission object.  This then results in an error with the API.

We should actually be catching this scenario during patching time, and providing users a helpful
error message letting them know that they should check their input parameters to `New-SubmissionPackage`
so that they know how to recover.

### DIFF
--- a/StoreBroker/StoreIngestionApplicationApi.ps1
+++ b/StoreBroker/StoreIngestionApplicationApi.ps1
@@ -1453,6 +1453,16 @@ function Patch-ApplicationSubmission
         }
     }
 
+    if (($AddPackages -or $ReplacePackages) -and ($NewSubmission.applicationPackages.Count -eq 0))
+    {
+        $output = @()
+        $output += "Your submission doesn't contain any packages, so you cannot Add or Replace packages." 
+        $output += "Please check your input settings to New-SubmissionPackage and ensure you're providing a value for AppxPath."
+        $output = $output -join [Environment]::NewLine
+        Write-Log $output -Level Error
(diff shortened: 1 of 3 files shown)


StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -1453,6 +1453,16 @@ function Patch-ApplicationSubmission
...
     }
 
+    if (($AddPackages -or $ReplacePackages) -and ($NewSubmission.applicationPackages.Count -eq 0))
+    {
+        $output = @()
+        $output += "Your submission doesn't contain any packages, so you cannot Add or Replace packages." 
+        $output += "Please check your input settings to New-SubmissionPackage and ensure you're providing a value for AppxPath."
+        $output = $output -join [Environment]::NewLine
+        Write-Log $output -Level Error
+        throw $output
```
danbelcher-MSFT: Can we do this with a helper method?

HowardWolosky: A helper method wouldn't buy us much here.  It's a conditional that throws a message.  The only thing that I could see realistically to share out would be the error message, but at this time, I don't see sufficient value for moving it to a common place when only being shared by these two instances.  Good suggestion, but won't fix.

Issue #28, opened 2017-03-03
Unable to submit common screenshot for all languages


DaveMebs:
My package has localized listing data, but the screenshots have nothing to localize - they are language/region agnostic. I'd like to be able to submit a common screenshot to be used everywhere, rather than creating required the folder structure per language and duplicating the file N times.

HowardWolosky:
@DaveMebs  - Sorry for not responding to this earlier.  I completely missed that this Issue was opened.  That's an interesting scenario.  So you're saying that even in your screenshot itself, there is no text or visual asset that would be localized/swapped out based on the user's language?

One way that I could see this being implemented would be to have a special "fallback" language folder during `Convert-ListingToObject` in `PackageTool.ps1`.  When it gets to `if ($null -eq $image)`, right now it errors out.  We could have one last check for that file in a special `common` folder (at the same level as the other languages) and it would look for the image there before failing.

That would be a fairly minimally invasive change and produce the result that you're looking for.  An additional change then to consider on top of that would be to copy it to a "common" language destination folder (as opposed to the language-specific sub-folder within the final StoreBroker payload) to reduce the size of the StoreBroker payload...otherwise the final payload will be larger than needed....we'd just have to check that the Submission API will allow shared screenshot files across languages (I'm not completely sure that it does).

HowardWolosky:
Hmmm...another option is to let users just specify an actual language to fallback to, and it would look there.  So, users could specify `en-us` as the fallback language, and it would look for UK screenshot images in `en-GB` first, and if not found, would then try grabbing them from the `en-us` fallback folder before failing altogether.  I feel like this is a more natural way of doing this.  Wondering if you have any specific thoughts @DaveMebs

DaveMebs:
In my specific case, I have a common image that is used across all languages with no localization ever used. Since the visual components of my package are not relevant to the user, I also only have this one image to upload.

In many ways, I like the idea of a fallback language over a fallback folder and agree with you that it seems more natural. At the same time, would someone ever have a screenshot localized in 50% of their target languages? That seems a bit odd and not a great user experience to design for.

The most likely case to me is that some images are localized and some are not. And the images that are not localized are never localized. It might make the most sense to have a "neutral" folder to support these and then consider enforcing that you cannot have an asset in both a lang folder and the neutral folder.

Thoughts?

HowardWolosky:
This was resolved a while ago with #80

PR #80, opened 2017-12-02
Enable Language Fallback support for media (screenshots and icons)

HowardWolosky:
The Store Submission API was recently updated to allow for submissions to reference the same media file across multiple elements.  In other words, the same .png file can now be referenced as a screenshot for multiple languages.

Some applications choose not to localize their screenshots, only focusing on localizing the written metadata for their PDP's.  In these scenarios, they would have a smaller StoreBroker package (and a simpler authoring environment) if they could simply tell StoreBroker to use a different language's screenshots when language-specific ones are not available.

With this change, the following end-user features are now enabled:
  * The StoreBroker config file has been updated to support a `MediaFallbackLanguage` config option
    that allows users to specify a lang-code (e,g, `en-us`, `es-es`, etc...)
  * Users can specify a `-MediaFallbackLanguage` parameter to `New-SubmissionPackage` and
    `New-InAppProductSubmissionPackage`, specifying a lang-code (e,g, `en-us`, `es-es`, etc...) which will
    override any config file option that has been specified.
  * Users can specify a `FallbackLanguage` attribute on the `<caption />` and `<icon />` elements in
     PDP files to optionally choose a different fallback language on a per-element basis.

In any of these cases, if the language-specific media file cannot be found, and a fallback language has
been specified, StoreBroker will then look in the fallback language's Images/media sub-folder for the
same-named media file.  Only then, if still not found, will packaging fail.

It should be noted that there can only ever be _one_ designated fallback language for any given file.
The logic does not cascade.  Therefore, if a user specifies a fallback language in both the config file
_and_ at the command prompt, the command prompt always wins.  If a value has been specified via
the config file and/or command prompt, _as well as_ in a PDP file, the fallback language from the PDP
file will be the one that is used _for that element_, and then StoreBroker will revert back to using the
one from the config file / command prompt until it encounters another one defined specifically in
the PDP file.

Addresses #28: Unable to submit common screenshot for all languages

### DIFF
--- a/PDP/ProductDescription.xsd
+++ b/PDP/ProductDescription.xsd
@@ -40,6 +40,12 @@
     </xs:restriction>
   </xs:simpleType>
 
+  <xs:simpleType name="LangCodeString">
+    <xs:restriction base="xs:normalizedString">
+      <xs:pattern value="\s*(\S(\r|\n|.){0,29}\s*)?"/>
+    </xs:restriction>
+  </xs:simpleType>
+
   <xs:attributeGroup name="ScreenshotImageGroup">
     <xs:attribute name="DesktopImage">
       <xs:simpleType>
@@ -76,6 +82,7 @@
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
+    <xs:attribute name="FallbackLanguage" type="tns:LangCodeString" />
   </xs:attributeGroup>
   
   <xs:simpleType name="CaptionStringContent">
@@ -137,7 +144,6 @@
   <xs:simpleType name="AppStoreNameString">
     <xs:restriction base="xs:string"/>
   </xs:simpleType>
-
 
   <xs:element name="ProductDescription">
     <xs:annotation>
@@ -245,6 +251,7 @@
                 </xs:complexType>
               </xs:element>
             </xs:sequence>
+            <xs:attribute name="FallbackLanguage" type="tns:LangCodeString" />
           </xs:complexType>
         </xs:element>
 
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -23,6 +23,7 @@ $script:s_AppxPath = "AppxPath"
 $script:s_OutPath = "OutPath"
 $script:s_OutName = "OutName"
 $script:s_DisableAutoPackageNameFormatting = "DisableAutoPackageNameFormatting"
+$script:s_MediaFallbackLanguage = "MediaFallbackLanguage"
 
 # String constants for application metadata
 $script:applicationMetadataProperties = @(
@@ -669,6 +670,14 @@ function Convert-ListingToObject
     .PARAMETER XmlFilePath
         A full path to the localized .xml file to be parsed.
         
+    .PARAMETER MediaFallbackLanguage
+        Some apps may not localize all of their metadata media (images, trailers, etc..)
+        across all languages.  By default, StoreBroker will look in the PDP langcode's subfolder
+        within ImagesRootPath for that language's media content.  If the requested filename is
+        not found, StoreBroker packaging will fail. If you specify a fallback language here
+        (e.g. 'en-us'), then if the requested file isn't found in the PDP language's media
+        subfolder, StoreBroker will then look into the fallback language's media subfolder for
+        the exactly same-named image, and only fail then if it still cannot be found.
 #>
     [CmdletBinding(SupportsShouldProcess)]
     param(
@@ -681,12 +690,15 @@ function Convert-ListingToObject
         [string[]] $LanguageExclude,
 
         [Parameter(Mandatory)]
+        [Alias('MediaRootPath')]
         [string] $ImagesRootPath,
 
         [Parameter(
             Mandatory,
             ValueFromPipeline)]
-        [string[]] $XmlFilePaths
+        [string[]] $XmlFilePaths,
+
+        [string] $MediaFallbackLanguage
     )
 
     PROCESS
@@ -775,13 +787,7 @@ function Convert-ListingToObject
                                                     ForEach-Object { $_.Trim() })
                     }           
 
-                    # ScreenshotCaption node is special, needs to be consumed separately
-                    $packageImagePath = Join-Path $script:tempFolderPath (Join-Path $script:packageImageFolderName $language)
-                    if (-not (Test-Path -PathType Container -Path $packageImagePath))
-                    {
-                        New-Item -ItemType directory -Path $packageImagePath | Out-Null
-                    }
-
+                    # Handle screenshots and their captions.
                     $imageListings = @()
                     $captions = $ProductDescriptionNode.ScreenshotCaptions.ChildNodes |
                         Where-Object NodeType -eq Element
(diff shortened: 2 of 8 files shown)


PDP/ProductDescription.xsd
```diff
@@ -76,6 +76,13 @@
         </xs:restriction>
       </xs:simpleType>
     </xs:attribute>
```
danbelcher-MSFT: So I'm thinking that, in most cases, if I want to specify a fallback language for a particular image, I probably want to specify the same fallback language for _all_ images in the set.  It would make most sense for this attribute to apply on the ScreenshotCaptions element.  If an individual Caption element has a different fallback language, then the language specified on the individual element should override the one specified on the ScreenshotCaptions element.

StoreBroker/PackageTool.ps1
```diff
@@ -681,12 +690,17 @@ function Convert-ListingToObject
...
 
         [Parameter(
             Mandatory,
             ValueFromPipeline)]
-        [string[]] $XmlFilePaths
+        [string[]] $XmlFilePaths,
+
+        [AllowNull()]
+        [AllowEmptyString()]
+        [string] $MediaFallbackLanguage
```
danbelcher-MSFT: No need for the Allow* attributes because the parameter is optional

StoreBroker/PackageTool.ps1
```diff
@@ -864,6 +860,86 @@ function Convert-ListingToObject
...
+    $mediaLanguageSourcePath = [System.IO.Path]::Combine($ImagesRootPath, $Release, $Language)
+
+    # The folder where we think the media file should be found if using the fallback language.
+    $mediaFallbackLanguageSourcePath = $null
+    if (-not [string]::IsNullOrEmpty($MediaFallbackLanguage) -and ($MediaFallbackLanguage -ne $Language))
+    {
+        $mediaFallbackLanguageSourcePath = [System.IO.Path]::Combine($ImagesRootPath, $Release, $MediaFallbackLanguage)
+        if (-not (Test-Path -Path $mediaFallbackLanguageSourcePath -PathType Container))
+        {
+            Write-Log "A fallback language was specified [$MediaFallbackLanguage], but a folder for that language does not exist [$mediaFallbackLanguageSourcePath], so media fallback support has been disabled." -Level Verbose
```
danbelcher-MSFT: I think we should error out in this case.  The user specified a fallback but the fallback does not exist, which indicates the state of their folders is not what the user intended.

StoreBroker/PackageTool.ps1
```diff
@@ -864,6 +860,86 @@ function Convert-ListingToObject
     }
 }
 
+function Get-LocalizedMediaFile
+{
+    [CmdletBinding(SupportsShouldProcess)]
```
danbelcher-MSFT: Needs CBH, especially a description that describes the steps used to resolve fallback image.

StoreBroker/PackageTool.ps1
```diff
@@ -864,6 +860,86 @@ function Convert-ListingToObject
...
+        if (-not (Test-Path -Path $mediaFallbackLanguageSourcePath -PathType Container))
+        {
+            Write-Log "A fallback language was specified [$MediaFallbackLanguage], but a folder for that language does not exist [$mediaFallbackLanguageSourcePath], so media fallback support has been disabled." -Level Verbose
+            $mediaFallbackLanguageSourcePath = $null
+        }
+    }
+
+    if (Test-Path -Path $mediaLanguageSourcePath -PathType Container)
+    {
+        $image = Get-ChildItem -Recurse -File -Path $mediaLanguageSourcePath -Include $Filename | Select-Object -First 1
```
danbelcher-MSFT: I'm not sure selecting the first one is the right choice here.  If there are multiple files with the same name under this path, I would rather error out so we don't accidentally make the wrong decision.  Worse yet, the user may not know that there we are selecting between multiple images and so it's harder for them to know that their structure is incorrect in the first place.

StoreBroker/PackageTool.ps1
```diff
@@ -775,13 +789,6 @@ function Convert-ListingToObject
...
                     }           
 
-                    # ScreenshotCaption node is special, needs to be consumed separately
-                    $packageImagePath = Join-Path $script:tempFolderPath (Join-Path $script:packageImageFolderName $language)
-                    if (-not (Test-Path -PathType Container -Path $packageImagePath))
-                    {
-                        New-Item -ItemType directory -Path $packageImagePath | Out-Null
-                    }
-
                     $imageListings = @()
```
danbelcher-MSFT: I'd like if we kept the comment about # ScreenshotCaptions because it tells the reader we're switching to handling the images now.

StoreBroker/PackageTool.ps1
```diff
@@ -801,39 +808,28 @@ function Convert-ListingToObject
...
-                                        "fileStatus"   = "PendingUpload";
-                                        "description"  = $caption.InnerText.Trim();
-                                        "imageType"    = $imageType;
-                                    }
+                                $params = @{
+                                    'Filename' = $imageFileName
+                                    'ImagesRootPath' = $ImagesRootPath
+                                    'Language' = $language
+                                    'Release' = $ProductDescriptionNode.Release
+                                    'MediaFallbackLanguage' = $requestedFallbackLanguage
```
danbelcher-MSFT: I don't think both Language and MediaFallbackLanguage should be parameters here.  If the user specified a MediaFallbackLanguage, then we should overwrite the value of $language and call Get-LocalizedMediaFile using the fallback language as the value of the parameter.  I'm not sure why Get-LocalizedMediaFile needs to know that any fallback language is being used at all.

StoreBroker/PackageTool.ps1
```diff
@@ -864,6 +860,86 @@ function Convert-ListingToObject
...
+        }
+    }
+
+    if (Test-Path -Path $mediaLanguageSourcePath -PathType Container)
+    {
+        $image = Get-ChildItem -Recurse -File -Path $mediaLanguageSourcePath -Include $Filename | Select-Object -First 1
+        $fileRelativePackagePath = [System.IO.Path]::Combine($script:packageImageFolderName, $Language, $Filename)
+    }
+
+    if (($null -eq $image) -and ($null -ne $mediaFallbackLanguageSourcePath))
```
danbelcher-MSFT: So I think the way this is written, it's covering for the case where the user specified a fallback language but there is also a matching image for the original language.  In this case, we're choosing the original image and ignoring what was specified for fallback language.  Again, I would rather error out here because it indicates that the user has an incorrect folder structure or an incorrect understanding of using fallback images.  I'd rather not make these decisions for them, instead just provide a helpful error message so they can fix the problem.  If packaging succeeds, it should be because we were clearly able to resolve packaging parameters without ambiguity.  This is just my approach so I'm assuming we'll talk more about it in-person.

PDP/ProductDescription.xsd
HowardWolosky: Agreed.  Added that as an option.

StoreBroker/PackageTool.ps1
HowardWolosky: Agreed

StoreBroker/PackageTool.ps1
HowardWolosky: Done

StoreBroker/PackageTool.ps1
HowardWolosky: Discussed in person.  This is a common helper designed to determine the correct file to use here.  As discussed in person, this is a _fallback_ feature, and not an _override_ feature.  We need both languages to be provided, because we need to check for the existence of the main language, and only then if not found will we then try finding the file in the fallback language's folder.

StoreBroker/PackageTool.ps1
HowardWolosky: Agreed. Oversight

StoreBroker/PackageTool.ps1
HowardWolosky: Disagreed.  The purpose of fallback is to default to prefer the language-specific one, and only use the fallback if the language-specific one can't be found. Won't fix.

StoreBroker/PackageTool.ps1
HowardWolosky: This was the pre-existing file selection logic.  I agree with your observation though and will improve this to error out if more than one matching file is found.

StoreBroker/PackageTool.ps1
HowardWolosky: Won't fix.  This is explicitly designed as a _fallback_ option, similar to how we fallback on the default language for localized strings if we can't find a specific version of that string for the current language.

PDP/ProductDescription.xsd
```diff
@@ -252,6 +252,13 @@
                 </xs:complexType>
               </xs:element>
             </xs:sequence>
+            <xs:attribute name="FallbackLanguage">
```
danbelcher-MSFT: Can't you declare the FallbackLanguage attribute once at the top of the file and then reference it both here and in the Caption element?

StoreBroker/PackageTool.ps1
```diff
@@ -925,6 +978,13 @@ function Get-LocalizedMediaFile
         throw $output
     }
 
+    if ($image.Count -gt 1)
+    {
+        $output = "More then one version of [$Filename] has been found for this language. Please ensure only one copy of this image exists within the language's sub-folders: [$($image -join ', ')]"
```
danbelcher-MSFT: The join operation needs $image.FullName

PDP/ProductDescription.xsd
HowardWolosky: Indeed.

StoreBroker/PackageTool.ps1
HowardWolosky: It actually worked fine without it (had verified before the update), but I agree that it's better to be declarative here.

PR #83, opened 2017-12-13
Fix an exception in the Update-*Submission cmdlets

danbelcher-MSFT:
Fix an exception in the Update-ApplicationSubmission, Update-ApplicationFlightSubmission, and Update-InAppProductSubmission that occurs when the user specifies a -SubmissionId but the submission is not in a 'PendingCommit' state.

While creating the exception message to throw, we forgot a '$' character and so PowerShell cannot interpret the sub-expression in the message.

### DIFF
--- a/StoreBroker/StoreBroker.psd1
+++ b/StoreBroker/StoreBroker.psd1
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.12.2'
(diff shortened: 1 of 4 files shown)


StoreBroker/StoreBroker.psd1
```diff
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.12.1'
+    ModuleVersion = '1.13.0'
```
HowardWolosky: This is a patch change as opposed to a minor change.

Issue #89, opened 2018-01-13
Assume neutral architecture for appx metadata when none specified


HowardWolosky:
Per the [MSDN documentation](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity), `neutral` is the default value for `ProcessorArchitecture`.

`Read-AppxMetadata` in `PackageTool.ps1` should be updated to default to `neutral` when it comes back with an empty or null value.

PR #90, opened 2018-01-13
Assume neutral architecture when none is specified

HowardWolosky:
Per the [MSDN documentation](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity),
`neutral` is the default value for `ProcessorArchitecture`.

Updating our Appx metadata reading logic to assume `neutral` when none is specified.

This will avoid a later exception in `Get-FormattedFilename` that does Metadata table validation.

Resolves Issue #89

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -1608,6 +1608,11 @@ function Read-AppxMetadata
 
         $metadata.version        = $manifest.Package.Identity.Version
         $metadata.architecture   = $manifest.Package.Identity.ProcessorArchitecture
+        if ([String]::IsNullOrWhiteSpace($metadata.architecture))
+        {
+            $metadata.architecture = "neutral"
+        }
+
         $metadata.targetPlatform = Get-TargetPlatform -AppxManifestPath $appxManifest
         $metadata.name           = $manifest.Package.Identity.Name -creplace '^Microsoft\.', ''
         
--- a/StoreBroker/StoreBroker.psd1
+++ b/StoreBroker/StoreBroker.psd1
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.12.3'
+    ModuleVersion = '1.12.4'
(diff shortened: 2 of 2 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1962,7 +1967,7 @@ function Get-FormattedFilename
     return $formattedBundleTags -join '_'
 }
 
-function Read-ApplicationMetadata
+function Read-OOoOOApplicationMetadata
```
danbelcher-MSFT: What happened here?

StoreBroker/PackageTool.ps1
```diff
@@ -1608,6 +1608,11 @@ function Read-AppxMetadata
 
         $metadata.version        = $manifest.Package.Identity.Version
         $metadata.architecture   = $manifest.Package.Identity.ProcessorArchitecture
+        if ([String]::IsNullOrWhiteSpace($metadata.architecture))
+        {
+            $metadata.architecture = "neutral"
```
danbelcher-MSFT: Could be single-assignment:
```
$metadata.architecture = if (...) { … } else { … }
```
Alternatively, I've been wanting a custom helper function like:
```
$metadata.architecture = If-NotNull -Then $manifest.Package.Identity.ProcessorArchitecture -Else "neutral"
```

StoreBroker/PackageTool.ps1
HowardWolosky: No idea.  Fixed, thanks.

StoreBroker/PackageTool.ps1
HowardWolosky: You're describing creating a [null-coalescing operator](https://en.wikipedia.org/w/index.php?title=Null_coalescing_operator&%3F_Operator=) ([`??`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operator) in C#). but for PowerShell.

Given that this is the type of base functionality that a language should support, and given the fact that it's [being considered for 6.1](https://github.com/PowerShell/PowerShell/issues/3240), I'd rather not try to roll our own at this time.

I'd prefer to leave as-is.

PR #94, opened 2018-01-20
Add auto-retry logic for configurable list of error codes

### LINKED ISSUE #92: Auto-Retry logic for 429 (API Rate Limit) errors
A tenant is limited to 20 requests per minute.
If that is exceeded, future requests will get back a `429` error until the next minute arrives.

We should update `Invoke-SBRestMethod` to auto-retry requests that fail due to `429` errors with an exponential fallback timeframe.  We should also consider randomizing the initial wait time to minimize the number of retries each instance may have to do.

HowardWolosky:
There is a set of known error codes where users can usually achieve success if they simply retry enough times:

  * `429` - The Submission API limits a Tenant to 20 requests per minute.  Any requests exceeding
       that will receive this error code until that minute has expired.

  * `503` - An underlying service component was unavailable, and the Submission API suggests that
       you try again.

This change modifies the default execution behavior of StoreBroker to auto-retry up to **5** times with exponential back-off for these two error codes.  The implementation adds two new global variables to permit users to further configure how this works:

 **`$global:SBAutoRetryErrorCodes`** - an `int` list of status codes that we should retry for
**`$global:SBMaxAutoRetries`** - an `int` indicating the maximum number of times we should retry before failing.

We will only report back telemetry when the retry count is reached (and we finally error out), or when the result is finally successful.  We are adding two new telemetry properties that will be reported for both (eventual) successful requests,, as well as for eventual exceptions:
   * `NumRetries` - to track the number of times we had to attempt a retry
   * `RetryStatusCode` - to track the last status code that was hit that caused us to need to do the retry.

Resolves #92


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1716,336 +1726,373 @@ function Invoke-SBRestMethod
...
-    # Add any special headers when using the proxy.
-    if ($serviceEndpoint -eq $script:proxyEndpoint)
+    
+    do
     {
-        if ($global:SBUseInt)
+        if ([System.String]::IsNullOrEmpty($AccessToken))
         {
-            $headers.Add("UseINT", "true")
+            $AccessToken = Get-AccessToken -NoStatus:$NoStatus
```
danbelcher-MSFT: What's the reasoning for this being in the loop?

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1716,336 +1726,373 @@ function Invoke-SBRestMethod
...
         }
+
+        # Since we have retry logic, we won't create a new stopwatch every time,
+        # we'll continue the existing one if it already exists...
```
danbelcher-MSFT: nit: I'd remove the 'if' because $stopwatch always exists

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: AccessTokens have a limited lifetime.  If we didn't get a new one each time, it's possible that it would be invalid during one of the retry attempts.

danbelcher-MSFT: Could you grab your comment and add it to the script?

HowardWolosky: Good feedback. Will do.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Agreed.  That comment was from a previous iteration

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -90,6 +90,16 @@ function Initialize-StoreIngestionApiGlobalVariables
     {
         $global:SBDefaultProxyEndpoint = $null
     }
+
+    if (!(Get-Variable -Name SBAutoRetryErrorCodes -Scope Global -ValueOnly -ErrorAction Ignore))
```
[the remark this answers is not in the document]
HowardWolosky: Interesting idea.  That _would_ work for most of them (immediate exception that I see is how `$SBLogPath` and `$SBLoggingEnabled` in `Helpers.ps1` have some interaction during initialization). I've opened #95 to track this suggestion for future consideration.

PR #97, opened 2018-01-24
Improve use of Write-Log and exceptions

danbelcher-MSFT:
Previously, users of Write-Log were left to handle creating messages themselves when logging an exception.  Typically, our use has been to extract the exception's message and log this string. The issue with this approach is that we lose other valuable information, such as the originating script, line, exception type, etc.

With this change, Write-Log now accepts a -Exception parameter and handles logging the full exception information.  This new functionality works in combination with the existing -Message parameter, so that users can log some context before logging the exception information.  We hope that this change will make debugging StoreBroker issues faster and more intuitive for the reader.

Additional changes:
- Write-Log properly handles pipelining multiple strings to the -Message parameter.
- Existing uses of Write-Log have been modified to use the new -Exception parameter where applicable.
- Spaces have been made uniform across touched files.

### DIFF
--- a/Extensions/ConvertFrom-ExistingIapSubmission.ps1
+++ b/Extensions/ConvertFrom-ExistingIapSubmission.ps1
@@ -94,7 +94,7 @@ function Add-ToElement
         PS C:\>Add-ToElement -Element $root -Comment "Comment1", "Comment2" -Attribute @{ "Attrib1"="Val1"; "Attrib2"="Val2" }
 
         Adds two comments and two attributes to the root element of the XML document.
-        
+
 #>
     param(
         [Parameter(Mandatory)]
@@ -268,7 +268,7 @@ function Add-Icon
         be included with a non-empty value.  This is fine if the Listing has an icon defined,
         but if it doesn't, then we want to include the element, but commented out so that users
         know later what they need to add if they wish to include an icon at some future time.
-        We will create/add the icon node the way we do in all other cases so that we have its XML, 
+        We will create/add the icon node the way we do in all other cases so that we have its XML,
         but if we then determine that there is no icon for that listing, we'll convert the element
         to its XML string representation that we can add as a comment, and then remove the actual
         node.
@@ -368,7 +368,7 @@ function ConvertFrom-Listing
         xml:lang="{0}"
         Release="{1}"/>', $Lang, $Release))
 
-    Add-Title -Xml $Xml -Listing $Listing    
+    Add-Title -Xml $Xml -Listing $Listing
     Add-Description -Xml $Xml -Listing $Listing
     $icon = Add-Icon -Xml $Xml -Listing $Listing
 
@@ -527,7 +527,7 @@ function Show-ImageFileNames
         {
             $output += " * [$langCode]: " + ($LangImageNames.$langCode -join ", ")
         }
-    
+
         Write-Log $($output -join [Environment]::NewLine)
     }
     else
@@ -580,7 +580,7 @@ function Main
             }
             catch
             {
-                Write-Log "Error creating [$lang] PDP: $($Error[0].Exception.Message )" -Level Error
+                Write-Log "Error creating [$lang] PDP:" -Exception $_ -Level Error
                 throw
             }
         }
--- a/Extensions/ConvertFrom-ExistingIapSubmission.ps1
+++ b/Extensions/ConvertFrom-ExistingIapSubmission.ps1
@@ -130,7 +130,7 @@ function Add-ToElement
         else
         {
             $out = "For element $($Element.LocalName), did not create attribute '$key' with value '$($Attribute[$key])' because the attribute already exists."
-            Write-Log $out -Level Warning
+            Write-Log -Message $out -Level Warning
         }
     }
 }
@@ -526,7 +526,7 @@ function Show-ImageFileNames
             $output += " * [$langCode]: " + ($LangImageNames.$langCode -join ", ")
         }
 
-        Write-Log $output
+        Write-Log -Message $output
     }
     else
     {
@@ -555,7 +555,7 @@ function Main
         if ([String]::IsNullOrEmpty($SubmissionId))
         {
             $SubmissionId = $iap.pendingInAppProductSubmission.id
-            Write-Log "No published submission exists for this In-App Product.  Using the current pending submission." -Level Warning
+            Write-Log -Message "No published submission exists for this In-App Product.  Using the current pending submission." -Level Warning
         }
     }
 
@@ -567,7 +567,7 @@ function Main
     $langs |
         ForEach-Object {
             $lang = $_.Name
-            Write-Log "Creating PDP for $lang" -Level Verbose
+            Write-Log -Message "Creating PDP for $lang" -Level Verbose
             Write-Progress -Activity "Generating PDP" -Status $lang -PercentComplete $(($pdpsGenerated / $langs.Count) * 100)
             try
             {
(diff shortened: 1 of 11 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -429,61 +429,129 @@ function Write-Log
...
         as well as to the default log file with the caller's username and a date/time stamp
         prepended to the message.
 
+    .EXAMPLE
+        try { $null.Do() }
+        catch { Write-Log -Message ("There was a problem.", "Here is the exception information:") -Exception $_ -Level Error }
+
+        Logs the message:
+
+        Write-LogHelper : 2018-01-23 12:57:37 : dabelc : There was a problem.
```
HowardWolosky: Not `Write-LogHelper` anymore, right?

StoreBroker/Helpers.ps1
```diff
@@ -495,19 +563,19 @@ function Write-Log
...
             $dateString,
             $env:username,
-            $Message
+            $finalMessage
 
         if ($global:SBShouldLogPid)
         {
-            $MaxPidDigits = 10 # This is an estimate (see https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows)
-            $pidColumnLength = $MaxPidDigits + "[]".Length
+            $maxPidDigits = 10 # This is an estimate (see https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows)
```
HowardWolosky: This had been capitalized to signify it as a constant

StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -1142,7 +1142,7 @@ function Update-ApplicationSubmission
                 $output = @()
                 $output += "We can only modify a submission that is in the '$script:keywordPendingCommit' state."
                 $output += "The submission that you requested to modify ($SubmissionId) is in '$($submissionToUpdate.status)' state."
-                
+
                 $newLineOutput = ($output -join [Environment]::NewLine)
```
HowardWolosky: It seems like there are probably a lot of opportunities throughout the code to now remove this type of line and just directly pass-in the array of strings, now that `Write-Log` will take care of NewLine'ing them for the user.

StoreBroker/Helpers.ps1
danbelcher-MSFT: Then I'd say it failed to signify that because I thought it was using the parameter naming convention.  Why not all caps?  Or local variable naming convention and use Set-Variable with the Constant/ReadOnly option to show that the value shouldn't be changed?

StoreBroker/Helpers.ps1
danbelcher-MSFT: Thanks

StoreBroker/StoreIngestionApplicationApi.ps1
danbelcher-MSFT: Sure.  I'll fix them.

StoreBroker/Helpers.ps1
HowardWolosky: At the time, I was unaware of the `ReadOnly`/`Constant` options for the `-Option` switch with `Set-Variable`....thanks for that.  PowerShell [best practices](https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/36) indicates PascalCase for constants.

StoreBroker/StoreIngestionApplicationApi.ps1
danbelcher-MSFT: Done

StoreBroker/Helpers.ps1
danbelcher-MSFT: 1. Revert this change.
2. Use Set-Variable with ReadOnly

Let me know which you'd prefer.

Extensions/ConvertFrom-ExistingIapSubmission.ps1
```diff
@@ -528,14 +526,13 @@ function Show-ImageFileNames
             $output += " * [$langCode]: " + ($LangImageNames.$langCode -join ", ")
         }
 
-        Write-Log $($output -join [Environment]::NewLine)
+        Write-Log $output
```
HowardWolosky: If you're here, I'd ask that you explicitly add the `-Message` at the same time, just to help us get up to the consistency we've been working towards.

Extensions/ConvertFrom-ExistingSubmission.ps1
```diff
@@ -1058,14 +1056,13 @@ function Show-ImageFileNames
             $output += " * [$langCode]: " + ($LangImageNames.$langCode -join ", ")
         }
 
-        Write-Log $($output -join [Environment]::NewLine)
+        Write-Log $output
```
HowardWolosky: same

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1377,7 +1377,7 @@ function Start-SubmissionMonitor
                     $shouldMonitor = $false
                 }
 
-                Write-Log $($body -join [Environment]::NewLine)
+                Write-Log $body
```
HowardWolosky: same

StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -300,7 +300,7 @@ function Format-Application
 
     End
     {
-       Write-Log $($output -join [Environment]::NewLine)
+       Write-Log $output
```
HowardWolosky: same

StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -524,7 +524,7 @@ function Format-ApplicationSubmission
 
     End
     {
-        Write-Log $($output -join [Environment]::NewLine)
+        Write-Log $output
```
HowardWolosky: same

StoreBroker/StoreIngestionFlightingApi.ps1
```diff
@@ -321,7 +321,7 @@ function Format-ApplicationFlight
 
     End
     {
-        Write-Log $($output -join [Environment]::NewLine)
+        Write-Log $output
```
HowardWolosky: same

StoreBroker/StoreIngestionFlightingApi.ps1
```diff
@@ -734,7 +733,7 @@ function Format-ApplicationFlightSubmission
...
```
HowardWolosky: same

StoreBroker/StoreIngestionIapApi.ps1
```diff
@@ -296,7 +296,7 @@ function Format-InAppProduct
 
     End
     {
-        Write-Log $($output -join [Environment]::NewLine)
+        Write-Log $output
```
HowardWolosky: same

StoreBroker/StoreIngestionIapApi.ps1
```diff
@@ -854,7 +854,7 @@ function Format-InAppProductSubmission
...
```
HowardWolosky: same

Extensions/ConvertFrom-ExistingIapSubmission.ps1
danbelcher-MSFT: I went all-in and replaced all instances of Write-Log that used implicit -Message parameter.

PR #101, opened 2018-02-02
Ensure that loc comments in PDP's are immediately before their content

HowardWolosky:
The Loc team's parser expects the loc comments to be directly before the content
that the comment refers to.

### DIFF
--- a/Extensions/ConvertFrom-ExistingIapSubmission.ps1
+++ b/Extensions/ConvertFrom-ExistingIapSubmission.ps1
@@ -206,7 +206,9 @@ function Add-Title
     $paramSet = @{
         "Element" = $elementNode;
         "Attribute" = @{ $script:LocIdAttribute = $script:LocIdFormat -f $elementName };
-        "Comment" = @($script:CommentFormat -f $maxChars, "IAP $elementName"; " [required] ")
+        "Comment" = @(
+            " [required] ",
+            ($script:CommentFormat -f $maxChars, "IAP $elementName"))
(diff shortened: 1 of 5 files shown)


Extensions/ConvertFrom-ExistingIapSubmission.ps1
```diff
@@ -206,7 +206,7 @@ function Add-Title
     $paramSet = @{
         "Element" = $elementNode;
         "Attribute" = @{ $script:LocIdAttribute = $script:LocIdFormat -f $elementName };
-        "Comment" = @($script:CommentFormat -f $maxChars, "IAP $elementName"; " [required] ")
+        "Comment" = @(" [required] "; $script:CommentFormat -f $maxChars, "IAP $elementName")
```
danbelcher-MSFT: Shouldn't this be a comma instead of semi-colon?

danbelcher-MSFT: Let's wrap the format operation in parentheses so we avoid the same situation we just ran into with Write-Log.  Here and elsewhere in this change (below)

HowardWolosky: Looks like both semi-colon _and_ comma are supported as array value separators...but I agree with you that we should be normalized on using comma.

HowardWolosky: agreed

PR #104, opened 2018-02-07
Better handling of proxy errors / prevent infinite retries when exceptions occur

HowardWolosky:
An exception that the Proxy might return will not contiain a `message` property
within the exception's `InnerMessage`, but the current error reporting logic
assumed that there would be a `message` property and then tried to call `Trim()`
on it.  This resulted in a `You cannot call a method on a null-valued expression.`
exception.

Because that exception occurred within a try/catch that was within a catch statement,
the outer catch ended up eating the exception, resulting in the retry loop continuing
to try the command, over and over again.

There are two distinct fixes in here as a result:

1. Only try to call `Trim()` if `InnerMessage.message` exists, otherwise,
   just report back `InnerMessage` itself.

2. Make sure that the outer catch statement ends with a `throw` to ensure
   that any inner exceptions thrown will propagate.  As a result, we will
   also explicitly `continue` the loop if we are going to perform an auto-retry.


StoreBroker/StoreBroker.psd1
```diff
@@ -6,7 +6,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '1.14.2'
+    ModuleVersion = '1.14.3'
```
danbelcher-MSFT: We're already at 1.14.3!

HowardWolosky: We weren't at the time that I pushed the PR.  Updated.

PR #105, opened 2018-02-08
Update CHANGELOG.md

HowardWolosky:
CHANGELOG.md hasn't been updated since StoreBroker was first released.
We should change that.
I've already gone through and added the appropriate [`git tags`](https://git-scm.com/docs/git-tag) for every release, and now the CHANGELOG will provide a very quick summary of every change as well for quick and easy perusal.

Going forward, we should keep both this CHANGELOG and the tags up-to-date.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,605 @@
 
 ------
 
+## [1.16.0](https://github.com/Microsoft/StoreBroker/tree/1.16.0) - (2018/02/14)
+### Features:
+
++ Add remaining localizable text fields, so that StoreBroker can now modify every aspect of a subimission (exposed by the API)
++ Support added for the following fields: `minimumHardware`, `shortDescription`, `shortTitle`, `sortTitle`, `voiceTitle`, `devStudio`
+
+More Info: [[pr]](https://github.com/Microsoft/StoreBroker/pull/108) | [[cl]](https://github.com/Microsoft/StoreBroker/commit/2dcce9c06e086b70a90e31144b57d8d3ecd9ad07)
+
+Author: [**@HowardWolosky**](https://github.com/HowardWolosky)
+
+------
+
+## [1.15.1](https://github.com/Microsoft/StoreBroker/tree/1.15.1) - (2018/02/12)
+### Features:
+
++ Add `-Force` switch to `Join-SubmissionPackage`, enabling you to overwrite an existing file
+
+More Info: [[pr]](https://github.com/Microsoft/StoreBroker/pull/106) | [[cl]](https://github.com/Microsoft/StoreBroker/commit/fb0764a4a14756fcd3161ca72630dafd5074063e)
+
+Author: [**@HowardWolosky**](https://github.com/HowardWolosky)
+
+------
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,608 @@
 
 ------
 
+## [1.15.1](https://github.com/Microsoft/StoreBroker/tree/1.15.1) - (2018/02/12)
```
danbelcher-MSFT: Seems like this is worth an update to CONTRIBUTING.md so others now how to update the changelog.

HowardWolosky: Done

CONTRIBUTING.md
```diff
@@ -305,6 +306,49 @@ When new code changes are checked in to the repo, a new NuGet package must be pu
 This process is documented in Microsoft's internal StoreBroker repo.
 Refer to the CONTRIBUTING.md in that repo for more information on creating a signed NuGet package.
 
+Once the new version has been pulled into master, there are two additional tasks to perform:
+  * Update [CHANGELOG.md](./CHANGELOG.md)
+  * Add a tag for that version to the repo
+
+### Updating the CHANGELOG
```
danbelcher-MSFT: These new sections are at the same level as the "Releasing" section but don't have a link from the ToC.  Should they be a subsection of "Releasing" or should they have links?

CONTRIBUTING.md
```diff
@@ -305,6 +306,49 @@ When new code changes are checked in to the repo, a new NuGet package must be pu
...
+To update [CHANGELOG.md](./CHANGELOG.md), just duplicate the previous section and update it to be
+relevant for the new release.  Be sure to update all of the sections:
+  * the version number
+  * the SB tree (we'll get that path working in a moment)
+  * the release date
+  * a brief list of all the changes (use a `-` for the bullet point if it's fixing a bug, or a `+` for a feature)
+  * the link to the pull request (pr) (so that the discussion on the change can be easily reviewed) and the changelist (cl)
+  * the author (and a link to their profile)
+  * If it's a new contributor, also add them to the [Contributors](#contributors) list below.
```
danbelcher-MSFT: nit: it's odd that all the other sentences in this list begin uncapitalized except for this one.

CONTRIBUTING.md
```diff
@@ -305,6 +306,49 @@ When new code changes are checked in to the repo, a new NuGet package must be pu
...
+
+Then get a new pull request out for that change to CHANGELOG.
```
danbelcher-MSFT: We should automate this

HowardWolosky: Great. Issue filed (#109).  Feel free to grab it if you're interested.  Won't block this change on it though.

CONTRIBUTING.md
HowardWolosky: changed

CONTRIBUTING.md
HowardWolosky: That was a mistake (thanks for catching).  They were supposed to be children of `Releasing`.  However, added to the TOC.

PR #110, opened 2018-02-15
Fix ConvertFrom-ExistingSubmission exception when no Advanced Listings support

HowardWolosky:
While this scenario had been tested brefore, clearly something had been missed.
It appears that if you pass `$null` (the result of $Submission.trailers` when
the application doesn't have Advanced Listing support) to `ForEach-Object`,
that iterates once (for the `$null`) which will eventually result into a call of
`$null | Get-Member -Type NoteProperty` which throws an exception since you haven't
passed an object to `Get-Member`.

The fix is to simply switch to a standard `foreach` as opposed to the `ForEach-Object`
cmdlet.

### DIFF
--- a/Extensions/ConvertFrom-ExistingSubmission.ps1
+++ b/Extensions/ConvertFrom-ExistingSubmission.ps1
@@ -933,72 +933,72 @@ function Add-Trailers
     Add-ToElement @paramSet
 
     $trailerCount = 0
-    $Submission.trailers |
-        ForEach-Object {
-            foreach ($language in ($_.trailerAssets | Get-Member -Type NoteProperty))
+    foreach ($trailer in $Submission.trailers)
+    {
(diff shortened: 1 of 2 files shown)


Extensions/ConvertFrom-ExistingSubmission.ps1
```diff
@@ -933,72 +933,72 @@ function Add-Trailers
     Add-ToElement @paramSet
 
     $trailerCount = 0
-    $Submission.trailers |
-        ForEach-Object {
```
danbelcher-MSFT: Alternatively, could have added a Where-Object { $null -ne $_ } |

HowardWolosky: Yeah, could have, but I feel like that would just add more questions as to "why would that happen?" as opposed to just using `foreach` where the behavior feels more clearly understood for anyone coming in to look at the code.

PR #112, opened 2018-04-04
Fix parsing of trailer titles when there is no inner comment

HowardWolosky:
The XML DOM that PowerShell returns treats inner text of an `XMLElement`
differently depending on whether it has any attributes/comments on it
vs just purely having text inside.

In the case of a `<Caption />` element, we know it will always have at
least one attribute on it.  That means that getting the caption text
will always be by querying for the `InnerText` of that element. The same
goes for the `<Image />` element for trailers: there must always be an
attribute specifying the image file, thus any text within the element
will always be accessed via `InnerText`.

For the `<Title />` element of a `<Trailer />`, things are different.
There are no attributes on that, so accessing the actual text of the
trailer will be different depending on whether the element has a
comment within the node or not.  At the moment, StoreBroker expected
there to always be that Loc comment within the `<Title />` element:

     

Thus, it always accessed the text via `InnerText.Trim()`.

A StoreBroker user ran into an issue when processing their trailers
when their localized PDP files did _not_ have this comment, as
packaging failed with this error:

     Convert-TrailersToObject : You cannot call a method on a null-valued expression.
     At StoreBroker\PackageTool.ps1:1432 char:9

In this scenario, calling `InnerText` returned `$null`, and so we were calling
`$null.Trim()` which resulted in the error.

The fix is to simply check if we're working with a String (the scenario when
there is no inner comment), and if not, to then proceed with getting the `InnerText`.
To ensure we don't have the problem anywhere else, I've made the fix more generic
with a helper method, and updated all of the places where we currently attempt
to access the text content of an `XMLElement`.

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -672,6 +672,62 @@ function Test-Xml
     }
 }
 
+function Get-TextFromXmlElement
+{
+    <#
+    .SYNOPSIS
+        Finds the actual text contained within an XMLElement, trims it and returns it.
+
+    .DESCRIPTION
+        Finds the actual text contained within an XMLElement, trims it and returns it.
+
+        When an XMLElement has attributes and/or inner comments, you cannot simply
+        access the text directly.  This function encapsulates the necessary logic
+        so that you can consistently and reliably get access to the actual element's
+        text.
+
+    .PARAMETER Element
+        This is the actual property from the XML DOM that you are intested in getting
+        the text of.  This might just be a String if it's a simple text element,
+        or it might end up being an XMLElement itself if has attributes and/or
+        comments.
+
+    .OUTPUTS
+        System.String.  The trimmed text contained within the XML element.
+
+    .EXAMPLE
+        Get-TextFromXmlElement -Element $xml.ProductDescription.Description
+
+        Returns back the trimmed text content contained within the Description element.
+#>
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory)]
+        [AllowNull()]
+        [AllowEmptyString()]
+        $Element
+    )
+
+    $text = $Element
+
+    # The text is only available to be accessed raw like this if there
+    # is no attribute or inner comment on the containing element.
+    # If there is, then this property will actually be an XmlElement
+    # itself and not a string, meaning that we have to access the
+    # InnerText to get its text value.
+    if ($text -is [System.Xml.XmlElement])
+    {
+        $text = $text.InnerText
+    }
+
+    if ($null -ne $text)
+    {
+        $text = $text.Trim()
+    }
+
+    return $text
+}
+
 function Convert-ListingToObject
 {
 <#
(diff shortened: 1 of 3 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1048,7 +1048,18 @@ function Convert-TrailersToObject
...
+                        $trailerTitle = $trailer.Title
+
+                        # The title text is only available to be accessed raw like that if there
+                        # is no attribute or inner comment.  If there is, then this Title property
+                        # will actually be an XmlElement itself and not a string, meaning that
+                        # we have to access the InnerText to get its text value.
+                        if ($trailerTitle.GetType() -eq [System.Xml.XmlElement])
+                        {
+                            $trailerTitle = $trailerTitle.InnerText
+                        }
```
danbelcher-MSFT: This fixes the issue locally but really we're at risk for any of the elements in the file.  For example, currently we process the Description element as a string, but a user could add a comment to that element and the DOM would start returning an XmlElement which would break us.  In general, we know the path to the content we want, we just aren't sure if that content will be represented as a String or an XmlElement.  We should write a generic function and convert all of our element processing to funnel through that function for retrieval of the content.  This may also be an opportunity for us to switch to XPath, if doing so solves this issue for us automatically.

StoreBroker/PackageTool.ps1
```diff
@@ -1048,7 +1048,18 @@ function Convert-TrailersToObject
...
                     foreach ($trailer in $trailers)
                     {
-                        $trailerTitle = $trailer.Title.InnerText.Trim()
+                        $trailerTitle = $trailer.Title
+
+                        # The title text is only available to be accessed raw like that if there
+                        # is no attribute or inner comment.  If there is, then this Title property
+                        # will actually be an XmlElement itself and not a string, meaning that
+                        # we have to access the InnerText to get its text value.
+                        if ($trailerTitle.GetType() -eq [System.Xml.XmlElement])
```
danbelcher-MSFT: Bonus points if you use built-in powershell features:
```
$trailerTitle -is [System.Xml.XmlElement]
```

HowardWolosky: Agreed. Done.  I had tried that in an earlier version and it wasn't working, but there must have been a different issue because this works fine now.

StoreBroker/PackageTool.ps1
HowardWolosky: We already had this solved correctly for all other properties in `Convert-ListingToObject` which checked both the property itself _as well as_ `InnerText`.  That being said, I think your suggestion to do this generically is a good idea, and I've done so in my latest rev.

StoreBroker/PackageTool.ps1
```diff
@@ -787,19 +843,14 @@ function Convert-ListingToObject
...
+                        Where-Object { ($null -ne (Get-TextFromXmlElement -Property $baseListing[$_])) }
 
                     $trimKeys | ForEach-Object {
-                        if ($null -ne $baseListing[$_].InnerText)
-                        {
-                            $baseListing[$_] = $baseListing[$_].InnerText
-                        }
-
-                        $baseListing[$_] = $baseListing[$_].Trim()
+                        $baseListing[$_] = Get-TextFromXmlElement -Property $baseListing[$_]
```
danbelcher-MSFT: There's still other places when processing the xml file where we reference InnerText.  We're still vulnerable to the same issue in those cases.  We need to update all our processing so that InnerText is only referenced from Get-TextFromXmlElement

HowardWolosky: Fixed as well, thanks.

PR #120, opened 2018-05-21
Cache the AccessToken for console session to speed up successive command requests

HowardWolosky:
Sped up the module for users not using a Proxy.  `AccessToken` is now cached for the
duration of the console session, and only needs to be refreshed when it expires (which is about
every 60 minutes).  Previously, the access token was only cached for the duration of the currently
executing command, which meant that any succesive interactions at the commandline required a new
`AccessToken` to be acquired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,24 @@
 
 ------
 
+## [1.17.0](https://github.com/Microsoft/StoreBroker/tree/1.17.0) - (2018/06/06)
+### Fixes:
+
+- Sped up the module for users not using a Proxy.  `AccessToken` is now cached for the
+  duration of the console session, and only needs to be refreshed when it expires (which is about
+  every 60 minutes).  Previously, the access token was only cached for the duration of the currently
+  executing command, which meant that any succesive interactions at the commandline required a new
+  `AccessToken` to be acquired.
+
+More Info: [[pr]](https://github.com/Microsoft/StoreBroker/pull/120) | [[cl]](https://github.com/Microsoft/StoreBroker/commit/TODO)
+
+Author: [**@HowardWolosky**](https://github.com/HowardWolosky)
+
+------
 ## [1.16.4](https://github.com/Microsoft/StoreBroker/tree/1.16.4) - (2018/04/23)
 ### Fixes:
 
-- Removed checks validation checks preventing some apps from updating
+- Removed checks validation checks which had prevented some apps from updating
   gaming options or trailers.  The API now enables all apps to use
   trailers and gaming options, even if their submission object doesn't
   provide those nodes.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,10 +3,24 @@
...
+
+Author: [**@HowardWolosky**](https://github.com/HowardWolosky)
+
+------
 ## [1.16.4](https://github.com/Microsoft/StoreBroker/tree/1.16.4) - (2018/04/23)
 ### Fixes:
 
-- Removed checks validation checks preventing some apps from updating
+- Removed checks validation checks prevented some apps from updating
   gaming options or trailers.  The API now enables all apps to use
```
danbelcher-MSFT: This sentence doesn't make much sense..

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -23,6 +23,13 @@ $script:accessTokenRefreshBufferSeconds = 90
 # do get an AccessToken so that it accurately reflects the time a token can last.
 $script:accessTokenTimeoutSeconds = (59 * 60) - $script:accessTokenRefreshBufferSeconds
 
+# We'll cache the last acquired acccess token so that we don't have to always get it
+# with every command within the same console session, provided that it hasn't expired.
+$script:lastAccessToken = $null
+
+# Indicates when $script:lastAccessToken has expired and must be refreshed
+$script:lastAccessTokenExpirationDate = Get-Date
```
danbelcher-MSFT: This works but  `[DateTime] 0` seems more like a `$null` equivalent.

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -295,6 +302,11 @@ function Set-StoreBrokerAuthentication
             $script:authCredential = $Credential
         }
     }
+
+    if ($PSCmdlet.ShouldProcess("", "Clear cached access token"))
+    {
+        $script:lastAccessToken = $null
```
danbelcher-MSFT: Do you need to reset `$script:lastAccessTokenExpirationDate`?  Maybe this logic should move to a `Clear-AccessToken` function?

CHANGELOG.md
HowardWolosky: agreed. fixed.

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: Interesting.  Wasn't aware of that.  Will leave this as-is however, as the code seems more clear/obvious if i'm saying "it expires right now"

StoreBroker/StoreIngestionApi.psm1
HowardWolosky: We only ever use the access token if it's non-null.  Will consider this feedback however.

HowardWolosky: Decided to won't fix this.

PR #124, opened 2018-08-07
Improve success rate for New-ApplicationSubmission

### LINKED ISSUE #122: Support isMinimalResponse when creating a new submission to reduce the load in the store backend
The store backend will support sending isMinimalResponse when a new submission is created. This will help the backend to reduce the load and increase the success rate of the creation of the submissions using the API.

HowardWolosky:
**Migrate `New-ApplicationSubmission` to use `isMinimalResponse`**
The Store API recently introduced a new option when cloning a new
Application submission (but not for IAP or Flight submissions):
    `isMinimalResponse=true`

When provided, the returned submission object will be shallow (missing
application packages and listings).  To get the full submission object,
you need to grab the `id` from that returned submission object and then
do a `GET` on that submission.

The benefit of doing this is reducing the likelihood that the API will
time out while trying to put together the newly cloned submission object
to return to us.  What has been happening is that when trying to clone
larger submissions, the service will return back a `500` error after it
has successfully cloned, but before it can return back the newly cloned
object.

**Fix conflicting access token caching mechanisms**
`Start-SubmissionMonitor` first added access-token caching logic to
limit the need for getting a new access token to just once per 58 minutes.

As of commit 35a9538, `Get-AccessToken` intelligently returns a cached
access token, only attempting to refresh it when it's about to expire.

Unfortunately, when paired together, an issue was introduced which meant that
`Start-SubmissionMonitor` might think that the _cached_ access token that it
got back had a `60 minute` expiration time (and assumed `58 minutes` in order
to have some additional buffer), but in reality, that access token might have
already been cached for 57 minutes.

Due to the double-caching logic, `Start-SubmissionMonitor` wouldn't attempt to
get a new access token until 58 minutes had passed, but by then the cached token
from `Get-AccessToken` may have already expired.

The fix here is to just let `Get-AccessToken` be the sole responsible entity for
caching tokens.  There's no reason to provide a "shared" access token to all
commands within `Start-SubmissionMonitor` for the same reason.   We'll just each
individual command try to acquire an AccessToken, letting `Get-AccessToken` do the
right thing in each instance.

**Fix `Format-ApplicationSubmission` check for trailers**
In the past, a submission object with no trailers had a `null` value
for the trailers node.  Now it appears that it returns back an empty
array instead.  Checkout `$trailers.Count -gt 0` handles both scenarios,
as `$null.Count -eq 0` is `$true` by default.

Resolves #122 Support isMinimalResponse when creating a new submission to reduce the load in the store backend

### DIFF
--- a/StoreBroker/StoreIngestionApplicationApi.ps1
+++ b/StoreBroker/StoreIngestionApplicationApi.ps1
@@ -542,7 +542,7 @@ function Format-ApplicationSubmission
             $output += ""
 
             # Only show the Trailers section if it exists
-            if ($null -ne $trailers)
+            if ($trailers.Count -gt 0)
             {
                 $langTrailers = $trailerByLang[$lang]
                 $output += "$(" " * $indentLength)Trailers            : {0}" -f $(if ($langTrailers.count -eq 0) { "<None>" } else { "" })
--- a/StoreBroker/StoreIngestionApplicationApi.ps1
+++ b/StoreBroker/StoreIngestionApplicationApi.ps1
@@ -873,7 +873,7 @@ function New-ApplicationSubmission
             }
         }
 
-        # Finally, we can POST with a null body to create a clone of the currently published submission
+        # Now we can POST with a null body to create a clone of the currently published submission
         $telemetryProperties = @{
             [StoreBrokerTelemetryProperty]::AppId = $AppId
             [StoreBrokerTelemetryProperty]::ExistingPackageRolloutAction = $ExistingPackageRolloutAction
@@ -881,7 +881,7 @@ function New-ApplicationSubmission
         }
 
         $params = @{
-            "UriFragment" = "applications/$AppId/submissions"
+            "UriFragment" = "applications/$AppId/submissions?isMinimalResponse=true"
             "Method" = "Post"
             "Description" = "Cloning current submission for App: $AppId"
             "AccessToken" = $AccessToken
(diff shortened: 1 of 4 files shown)


StoreBroker/StoreIngestionApplicationApi.ps1
```diff
@@ -873,15 +873,15 @@ function New-ApplicationSubmission
...
+        # Now we can POST with a null body to create a clone of the currently published submission
         $telemetryProperties = @{
             [StoreBrokerTelemetryProperty]::AppId = $AppId
             [StoreBrokerTelemetryProperty]::ExistingPackageRolloutAction = $ExistingPackageRolloutAction
             [StoreBrokerTelemetryProperty]::Force = $Force
         }
 
         $params = @{
-            "UriFragment" = "applications/$AppId/submissions"
+            "UriFragment" = "applications/$AppId/submissions?isMinimalResponse=true"
```
danbelcher-MSFT: My one concern with using this, is that it's not listed in the docs: https://docs.microsoft.com/en-us/windows/uwp/monetize/create-an-app-submission

danbelcher-MSFT: Is this not available for Flights and Iap?

HowardWolosky: I'll follow-up with them to see if they're planning to update MSDN with this parameter.  They've specifically asked me to adopt this change however, so I'm not too concerned about adopting it without an MSDN update.

HowardWolosky: Correct. They indicated that getting it working for flights and IAP's would be quite a bit more work.

PR #134, opened 2018-09-07
StoreBrokerProxy should proxy the Location header from the real response

HowardWolosky:
For API v2, the API will start returning a Location header that clients
should honor for subsequent requests.  This follows the
[HTTP/1.1 standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5)
which states:

>if a `POST` operations results in the creation of a resource, then the reponse
> should include a `Location` header with the address of the new resource.

The API will be using the `Location` for other scenarios as well, but in order for
StoreBroker to leverage it, we need to ensure that the Proxy is returning the original
value.


RESTProxy/Models/Endpoint.cs
```diff
@@ -370,6 +371,10 @@ private HttpResponseMessage GetResponseMessage(HttpWebResponse httpResponseFromA
                 {
                     httpResponseMessage.Headers.Add(key, httpResponseFromApi.Headers[key]);
                 }
+                else if (key.ToLowerInvariant() == LocationHeader.ToLowerInvariant())
+                {
+                    httpResponseMessage.Headers.Add(key, httpResponseFromApi.Headers[key]);
```
danbelcher-MSFT: You can reduce this sequence of `if-else`s to a single conditional because the action is always the same.

PR #138, opened 2018-10-09
Protect against formatting invalid certification reports

HowardWolosky:
Per the [documentation](https://docs.microsoft.com/en-us/windows/uwp/monetize/manage-app-submissions#certification-report-object),
a certification report is supposed to have both a `date` and `reportUrl` properties.

We've seen instances where the API returns back certification reports that don't have a `date` property, and where the `reportUrl`
contains the message `Obsolete: please visit https://developer.microsoft.com for report details.`.  When this happens, we end up
throwing an exception as we attempt to format a `$null` date value.

This update protects against that scenario by ensuring that we only attempt to format the date if one is provided.

### DIFF
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -1344,7 +1344,13 @@ function Start-SubmissionMonitor
                 $body += "Status Details [Certification Reports] : {0}" -f $(if ($submission.statusDetails.certificationReports.count -eq 0) { "<None>" } else { "" })
                 foreach ($report in $submission.statusDetails.certificationReports)
                 {
-                    $body += $(" " * $indentLength) + $(Get-Date -Date $report.date -Format R) + ": $($report.reportUrl)"
+                    $date = ""
+                    if ($null -ne $report.date)
+                    {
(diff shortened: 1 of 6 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -1344,7 +1344,13 @@ function Start-SubmissionMonitor
                 $body += "Status Details [Certification Reports] : {0}" -f $(if ($submission.statusDetails.certificationReports.count -eq 0) { "<None>" } else { "" })
                 foreach ($report in $submission.statusDetails.certificationReports)
                 {
-                    $body += $(" " * $indentLength) + $(Get-Date -Date $report.date -Format R) + ": $($report.reportUrl)"
+                    $date = ""
+                    if ($null -ne $report.date)
```
danbelcher-MSFT: If we're going to repeat this logic everywhere, maybe it's worth having a safe form of Get-Date in the helpers?

HowardWolosky:
Abandoning this change.  After contacting the API team, they are adding the `date` field (along with a valid value) back to the `certificationReports` result, so this is no longer necessary.

PR #150, opened 2019-03-15
No Status switch doesn't cache the access token

elendil326:
Currently, the method `Get-AccessToken` only caches the access token when the `NoStatus` switch is not provided. The caching mechanism should be used regardless of the `NoStatus` flag to avoid multiple calls to the OAuth API.

### DIFF
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -522,8 +522,6 @@ function Get-AccessToken
             {
                 $response = Invoke-RestMethod $url -Method Post -Body $body
             }
-
-            return $response.access_token
         }
         else
         {
@@ -553,14 +551,20 @@ function Get-AccessToken
             {
                throw $remoteErrors[0].Exception
             }
+        }
 
+        # Account for the case where ShouldProcess is false
+        if (Test-Path variable:response)
+        {
             # Keep track of how long this token will be valid for, to enable logic that re-uses
             # the same token across multiple commands to know when a new one is necessary.
             $script:accessTokenTimeoutSeconds = $response.expires_in - $script:accessTokenRefreshBufferSeconds
             $script:lastAccessTokenExpirationDate = (Get-Date).AddSeconds($script:accessTokenTimeoutSeconds)
             $script:lastAccessToken = $response.access_token
-            return $response.access_token
+            Write-Log -Message "Access Token has been cached for future use. Will expire in $($script:accessTokenTimeoutSeconds) seconds."
         }
+
+        return $response.access_token
     }
     catch [System.InvalidOperationException]
     {
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -515,6 +515,7 @@ function Get-AccessToken
     {
         Write-Log -Message "Getting access token..." -Level Verbose
         Write-Log -Message "Accessing [POST] $url" -Level Verbose
+        $response = $null
 
         if ($NoStatus)
         {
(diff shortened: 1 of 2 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -553,14 +551,20 @@ function Get-AccessToken
...
+        # Account for the case where ShouldProcess is false
+        if (Test-Path variable:response)
+        {
             # Keep track of how long this token will be valid for, to enable logic that re-uses
             # the same token across multiple commands to know when a new one is necessary.
             $script:accessTokenTimeoutSeconds = $response.expires_in - $script:accessTokenRefreshBufferSeconds
             $script:lastAccessTokenExpirationDate = (Get-Date).AddSeconds($script:accessTokenTimeoutSeconds)
             $script:lastAccessToken = $response.access_token
-            return $response.access_token
+            Write-Log -Message "Access Token has been cached for future use. Will expire in $($script:accessTokenTimeoutSeconds) seconds."
```
HowardWolosky: We want to avoid writing anything directly to the screen unless it's truly relevant.  I'm fine with logging this, but it should be `-Level Verbose`.

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -553,14 +551,20 @@ function Get-AccessToken
             {
                throw $remoteErrors[0].Exception
             }
+        }
 
+        # Account for the case where ShouldProcess is false
+        if (Test-Path variable:response)
```
HowardWolosky: This one is cofusing me.  I'm not sure what `variable:response` is, nor what you're using `Test-Path` for with it.  Based on your comment, if your intent is to protect this around `ShouldProcess`, you'd want to use `if ($PSCmdlet.ShouldProcess("Cache the access token"))`

StoreBroker/StoreIngestionApi.psm1
```diff
@@ -553,14 +551,20 @@ function Get-AccessToken
...
             # Keep track of how long this token will be valid for, to enable logic that re-uses
             # the same token across multiple commands to know when a new one is necessary.
             $script:accessTokenTimeoutSeconds = $response.expires_in - $script:accessTokenRefreshBufferSeconds
             $script:lastAccessTokenExpirationDate = (Get-Date).AddSeconds($script:accessTokenTimeoutSeconds)
             $script:lastAccessToken = $response.access_token
-            return $response.access_token
+            Write-Log -Message "Access Token has been cached for future use. Will expire in $($script:accessTokenTimeoutSeconds) seconds."
         }
+
+        return $response.access_token
```
HowardWolosky: Since you've changed the scope of where you're accessing `$response`, can you please do an initial define of `$response = $null` at https://github.com/Microsoft/StoreBroker/blob/d95aea15a2e10e05e553cf5d3e11797972f6335d/StoreBroker/StoreIngestionApi.psm1#L518?

StoreBroker/StoreIngestionApi.psm1
elendil326: That is a way to verify whether a variable is defined before accessing it. It is important when running under `Strict-Mode -Version 2.0`. In this case, the response variable will not exist if the user decided not to process the `Invoke-RestMethod` action. We could verify it under another `ShouldProcess` as you suggest but I don't know how that would behave if the user decided not to process the `Invoke-RestMethod` and decides to process `Cache the access token`.

HowardWolosky: Ah, cool.  That's new to me.  I'm fine with leaving it as `Test-Path` then.  Just please explicitly add `-Path` before `variable:response`, and it would be great if you could just add a comment about what it's doing there, given that we don't have any other current examples of that anywhere else in the codebase.

HowardWolosky: Ah right, if you define it at the top of the `try`, then this check would no longer be valid.  Your new update makes more sense now, thanks.

Issue #179, opened 2019-10-29
New-StoreBrokerConfigFile generates incorrect targetPublishDate value in PowerShell Core 6.2+


riverar:
Windows 10 1909 18363.418
PSVersion 6.2.0 ⚠ PowerShell Core ⚠ (Also seeing this on Azure DevOps with Windows PowerShell)
StoreBroker 1.19.1

When generating a new configuration via `New-StoreBrokerConfigFile`, the `targetPublishDate` value is incorrectly specified as an object containing `value` and `DateTime` properties. When attempting to submit this later via `Update-ApplicationSubmission` the API will reject it.

Generated snippet:
```json
"targetPublishDate": {
  "value": "1601-01-01T00:00:00Z",
  "DateTime": "Monday, January 1, 1601 12:00:00 AM"
}
```

Sample response (`MS-CorrelationId: e767cde6-b049-4908-804f-beeaa36eafbd`):
```json
{
  "code": "InvalidParameterValue",
  "data": [],
  "details": [],
  "message": "Unexpected character encountered while parsing value: {. Path 'targetPublishDate', line 12, position 24.",
  "source": "Ingestion Api",
  "target": "submissionRequest.targetPublishDate"
}
```

riverar:
Thought this issue was specific to PowerShell Core but am also seeing this on Azure DevOps using Windows PowerShell 5.x.

HowardWolosky:
Interesting.  This does not repro for me at all on PS 5.1.19013.1002.

Here's the simple repro:

```powershell
  $rawSub = '{"id":"123245","applicationCategory":"UtilitiesAndTools","visibility":"Public","targetPublishMode":"Immediate","targetPublishDate":"1601-01-01T00:00:00.0000000Z","friendlyName":"Submission 87","trailers":[]}'
  $sub = $rawSub | convertfrom-json
  Write-Host ('          "targetPublishDate": .*,' -replace '"targetPublishDate": .*,', "`"targetPublishDate`": $($sub.targetPublishDate | ConvertTo-Json)")
```

On PS 5.1.19013.1002, this returns back:
`          "targetPublishDate": "1601-01-01T00:00:00.0000000Z"`

I'd like to better understand why we're seeing a difference in behavior here before I'd take any change to the code.

riverar:
Yep, looks to be PowerShell Core specific.

PowerShell Core 6.2.0
```powershell
[...]

PR #180, opened 2019-10-29
Explicitly coerce TargetPublishDate into ISO8601 prior to use in json

riverar:
Resolves https://github.com/microsoft/StoreBroker/issues/179

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -112,7 +112,7 @@ function Get-StoreBrokerConfigFileContentForIapId
 
         # PUBLISH MODE AND VISIBILITY
         $updated = $updated -replace '"targetPublishMode": ".*",', "`"targetPublishMode`": `"$($sub.targetPublishMode)`","
-        $updated = $updated -replace '"targetPublishDate": .*,', "`"targetPublishDate`": `"$($sub.targetPublishDate)`","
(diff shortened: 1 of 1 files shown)


HowardWolosky:
Per the comments I added in #179, I'm not able to repro this problem.  Before I'd take a change to the code, I'd like to better understand why we're seeing a difference in behavior.  I'm going to close this PR down for now and we can continue the discussion via #179 until such time as we understand why you're seeing this problem and we're not.

riverar:
@HowardWolosky If you're starting to ramp up PowerShell Core support, you may want to revisit this PR. 

Per https://github.com/microsoft/StoreBroker/issues/179#issuecomment-547650786, I suspect you're going to run into some date/time issues soon.

HowardWolosky:
Thanks for reminding me about this, @riverar. At the time that you brought up these issues, I didn't have a good understanding of PowerShell Core and its future/direction.  I've been working with it quite a bit over in [PowerShellForGitHub](https://github.com/microsoft/PowerShellForGitHub) lately, so now I just have my head wrapped around it better.  Especially with the advent of GitHub Actions (which use PS Core), it makes a lot more sense now to make sure that this is working in all PS environments.

riverar:
Just got the alert on this, oddly. Will get this done by EOB.

HowardWolosky:
> Will get this done by EOB.

Thanks so much!

PR #185, opened 2020-01-21
User/lukec/storeUrls

cartwrightluke:
Updated all of the printed storebroker URLs to use the new url format:
dev.windows -> developer.microsoft -> partner.microsoft
iaps - > products

Previously, these links were broken.


cartwrightluke:
Thanks! I've updated the version number, and made a new pr into the V2 branch.

PR #187, opened 2020-02-19
Don't overwrite inner package metadata with data from packages in subfolders

joseartrivera:
Packages in subfolders should have a subset of capabilities of the main app. It should also have the same target device family and platform. Furthermore we shouldn't use the version number from the subfolder package when naming the package later on.

### DIFF
--- a/StoreBroker/PackageTool.ps1
+++ b/StoreBroker/PackageTool.ps1
@@ -1891,16 +1891,16 @@ function Read-AppPackageBundleMetadata
         $targetDeviceFamilies            = @()
         $targetDeviceFamiliesEx          = @()
 
-        $applications = ($manifest.Bundle.Packages.ChildNodes | Where-Object Type -like "application").FileName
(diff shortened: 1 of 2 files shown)


StoreBroker/PackageTool.ps1
```diff
@@ -1909,6 +1909,12 @@ function Read-AppPackageBundleMetadata
             $capabilities            += $appPackageMetadata.capabilities
             $targetDeviceFamilies    += $appPackageMetadata.targetDeviceFamilies
             $targetDeviceFamiliesEx  += $appPackageMetadata.targetDeviceFamiliesEx
+            
+            # Don't overwrite full app data with data from a stub
+            if ($application.IsStub -and $null -ne $metadata.innerPackages.$($appPackageMetadata.architecture))
```
HowardWolosky:
```suggestion
            if ($application.IsStub -and ($null -ne $metadata.innerPackages.$($appPackageMetadata.architecture)))
```

In general, the convention in this project is to wrap all expressions in parenthesis.

PR #189, opened 2020-03-20
Make the transfer limit used for AzSDML configurable.

HowardWolosky:
AzSDML = Azure Storage Data Movement Library

For users that run multiple pararallel StoreBroker clients,
it's possible that they may run out of system resources when
those clients all start trying to tranfer packages and/or
media metadata (screenshots/videos) concurrently.

This change allows users to adjust the default connection limit that the Azure Storage Data Movement Library uses when
transferring content.

### DIFF
--- a/StoreBroker/StoreIngestionApi.psm1
+++ b/StoreBroker/StoreIngestionApi.psm1
@@ -151,6 +151,11 @@ function Initialize-StoreIngestionApiGlobalVariables
     {
         $global:SBStoreBrokerClientName = $null
     }
+
+    if (!(Get-Variable -Name SBDefaultTransferConnectionLimit -Scope Global -ValueOnly -ErrorAction Ignore))
+    {
+        $global:SBDefaultTransferConnectionLimit = [Environment]::ProcessorCount * 8
+    }
 }
 
 # We need to be sure to call this explicitly so that the global variables get initialized.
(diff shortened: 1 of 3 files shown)


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -905,10 +910,10 @@ function Set-StoreFile
             if ($PSCmdlet.ShouldProcess($jobName, "Start-Job"))
             {
                 [scriptblock]$scriptBlock = {
-                    param($SasUri, $FilePath, $AzureStorageDll, $AzureStorageDataMovementDll)
+                    param($SasUri, $FilePath, $AzureStorageDll, $AzureStorageDataMovementDll, $DefaultTransferConnectionLimit)
 
                     # Recommendations per https://github.com/Azure/azure-storage-net-data-movement#best-practice
-                    [System.Net.ServicePointManager]::DefaultConnectionLimit = [Environment]::ProcessorCount * 8
+                    [System.Net.ServicePointManager]::DefaultConnectionLimit = $DefaultTransferConnectionLimit
```
joseartrivera: I wonder how this correlates to a similar option in the Microsoft.WindowsAzure.Storage.DataMovement.TransferManager namespace called parallel operations.

[https://docs.microsoft.com/en-us/previous-versions/azure/reference/mt685063%28v%3dazure.100%29](https://docs.microsoft.com/en-us/previous-versions/azure/reference/mt685063%28v%3dazure.100%29)

Maybe we should set that too? Or have an option to modify it?

HowardWolosky: Interesting.  I'd only want to do that if I could provide adequate documentation for what the difference is between the two.

I found this [GitHub issue](https://github.com/Azure/azure-storage-net-data-movement/issues/23) where they were discussing what the default values of each should be, but it still doesn't give me a clear understanding of how to express to users why they would set one value vs the other.

HowardWolosky: I gave it a try. `[Microsoft.WindowsAzure.Storage.DataMovement]::ParallelOperations` wasn't accessible (specifically, `Unable to find type [Microsoft.WindowsAzure.Storage.DataMovement].`)  For the time being, I'm fine with just exposing `DefaultConnectionLimit`.

joseartrivera: Thanks for investigating! Sounds good to me

PR #193, opened 2020-06-05
Ensure PSCustomObjects correctly have their enums converted to strings

HowardWolosky:
> Note, there's something wrong with this branch based on the commits I'm seeing below (it should have just been directly on top of v2@head), so please don't merge this PR in.  I'm posting it for reference right now.

StoreBroker uses custom Enumerations in order to prevent typos for certain
values.  A side-effect of that, is that we have to do extra work when
converting objects to JSON because ConvertTo-Json will by default
convert an enumeration to its _numerical value_ instead of its _name_.

That's why we use Get-JsonBody, which internally uses ConvertTo-Json after
calling Convert-EnumToString to ensure that all enums have been converted
to their string names first.

The problem we had here was that if the object being passed in was a
PSCustomObject, Convert-EnumToString was leaving everything as-is.
This became a problem in a scenario like New-Submission, which calls
Get-SubmissionRollout, modifies the `state` property, and then calls
Set-SubmissionRollout with that modified object.  That modified object
is a PSCustomObject, but its state is now an Enum value which needs
to be converted back into a string.

There are two potential fixes here:
 1. For all known scenarios where this might occur, append `.ToString()`
    to the enum assignment.  This leaves open the possibility that some
    existing cases might be missed, or that new cases may be introduced
    in the future.
 2. Update Convert-EnumToString to understand PSCustomObjects and do
    the right thing.

This change takes the second approach.


StoreBroker/Helpers.ps1
```diff
@@ -1242,6 +1242,21 @@ function Convert-EnumToString
...
         return $modified
     }
+    elseif ($InputObject -is [PSCustomObject])
+    {
+        $modified = $InputObject.PSObject.Copy() # Get a new instance, not a reference
+        $modified | Get-Member -type NoteProperty | Select-Object -ExpandProperty Name | ForEach-Object {
+            $converted = (Convert-EnumToString -InputObject $modified.$_)
+            if ($modified.$_ -is [array])
+            {
+                $converted = @(converted)
```
joseartrivera, line 1252: When running this locally with an array inside of a PSCustomObject I get the below error and the array appears to be empty.

```
converted : The term 'converted' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify 
that the path is correct and try again.
At C:\StoreBroker\StoreBroker\Helpers.ps1:1252 char:32
+                 $converted = @(converted)
+                                ~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (converted:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

PScustomObject  created like this:
```
$test = [pscustomobject]@{
>> state = [StoreBrokerRolloutState]::RolledBack
>> list = @([StoreBrokerRolloutState]::RolledBack, [StoreBrokerRolloutState]::Completed)
>> }
```

Should it be this?
```suggestion
                $converted = @($converted)
```

HowardWolosky: Indeed.  I had tested this before submitting, so I must have done a ctrl-z/ctrl-s at some point after doing the import that verified.  Thanks for re-verifying and catching the typo.

PR #194, opened 2020-06-29
Check for null values when calling Convert-EnumToString on PSCustomObject

joseartrivera:
Convert-EnumToString iterates over each property and calls itself recursively in order to convert all properties that are enums into strings. In the case of converting PSCustomObjects if the value in the property is $null we would try to pass that $null value recursively to Convert-EnumToString which would throw this exception:

Cannot bind argument to parameter 'InputObject' because it is null.

Now we check for null before doing the recursive call.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -1246,13 +1246,16 @@ function Convert-EnumToString
     {
         $modified = $InputObject.PSObject.Copy() # Get a new instance, not a reference
         $modified | Get-Member -type NoteProperty | Select-Object -ExpandProperty Name | ForEach-Object {
-            $converted = (Convert-EnumToString -InputObject $modified.$_)
-            if ($modified.$_ -is [array])
+            if ($null -ne $modified.$_)
             {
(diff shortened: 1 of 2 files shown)


joseartrivera:
> Nice catch.
> An alternate solution would be to add `[AllowNull()]` on the `$InputObject` in the `params` definition, but this works as well.
> 
> Something to consider: Line 1240 could also conceivably error out if a hashtable entry has a `$null` value (because of that `$key.ToString()` call)... If you're shoring up this function for possible `$null` values, you should probably guard there too.

Good catch and good idea, I like your approach better since it covers both cases.

StoreBroker/Helpers.ps1
```diff
@@ -1211,6 +1211,7 @@ function Convert-EnumToString
         [Parameter(
             ValueFromPipeline,
             Mandatory)]
+        [AllowNull()]
```
HowardWolosky, line 1214: Interesting.  I was about to argue with you that this doesn't solve the second scenario that I brought up...but it does (I hadn't realized that).

PR #213, opened 2020-11-09
Only resolve MediaRootPath if it's necessary

HowardWolosky:
When `Update-Listing` calls into `Update-ListingImage` to remove unneeded images, it calls that with `-RemoveOnly` and doesn't bother passing in a `MediaRootPath`.  However, we still tried to resolve the `MediaRootPath`, and that could cause an exception since the path being resolved was blank.

https://github.com/microsoft/StoreBroker/blob/6e70140246592a37a9156a1c4000b87d37de7055/StoreBroker/StoreIngestionListingApi.ps1#L827-L843

https://github.com/microsoft/StoreBroker/blob/6e70140246592a37a9156a1c4000b87d37de7055/StoreBroker/StoreIngestionListingImageApi.ps1#L465

This updates `Update-ListImage` to only resolve `MediaRootPath` if we're going to use it (meaning that `RemoveOnly` wasn't specified).

### DIFF
--- a/StoreBroker/StoreIngestionListingImageApi.ps1
+++ b/StoreBroker/StoreIngestionListingImageApi.ps1
@@ -462,8 +462,6 @@ function Update-ListingImage
     {
         $stopwatch = [System.Diagnostics.Stopwatch]::StartNew()
 
-        $MediaRootPath = Resolve-UnverifiedPath -Path $MediaRootPath
-
         $params = @{
             'ProductId' = $ProductId
             'SubmissionId' = $SubmissionId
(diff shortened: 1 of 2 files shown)


StoreBroker/StoreIngestionListingImageApi.ps1
```diff
@@ -462,8 +462,6 @@ function Update-ListingImage
     {
         $stopwatch = [System.Diagnostics.Stopwatch]::StartNew()
 
-        $MediaRootPath = Resolve-UnverifiedPath -Path $MediaRootPath
```
joseartrivera, line 465: Will the reference in telemetry cause any issues?

Get-PiiSafeString -PlainText $MediaRootPath

HowardWolosky: Good question.
No, it won't.

`Get-PiiSafeString` allows for receiving an empty string to protect
https://github.com/microsoft/StoreBroker/blob/49e289e1c10e581fe5638b1be592a7b131acb96a/StoreBroker/Telemetry.ps1#L149-L154

(as does `Get-SHA512Hash` which it calls internally)
https://github.com/microsoft/StoreBroker/blob/49e289e1c10e581fe5638b1be592a7b131acb96a/StoreBroker/Helpers.ps1#L312-L317

They just return back an empty string.

As for telemetry, it's perfectly valid to send a telemetry property up that has no value.  So...it shouldn't cause any problem.

PR #214, opened 2020-11-13
Fix issues when removing a listing from a new submission

joseartrivera:
When a new submission is cloned that does not include a listing for a language that was the previously included, the code path to removing the listing for this language has a couple of issues:

- We delete the listing first, and then we try to remove the images/videos. This results in a 404 error in the latter API calls since the listing has already been removed at that point. We should remove the images/videos first and then do the API call to delete the entire listing.
- We try to verify $MediaRootPath when calling Update-ListingVideo with the -RemoveOnly flag causing an error since $MediaRootPath would be null in this scenario

### DIFF
--- a/StoreBroker/StoreIngestionListingApi.ps1
+++ b/StoreBroker/StoreIngestionListingApi.ps1
@@ -836,9 +836,9 @@ function Update-Listing
             Write-Log -Message 'Now removing listings for languages that were cloned by the submission but don''t have current user data.' -Level Verbose
             foreach ($langCode in $listingsToDelete)
             {
-                $null = Remove-Listing @commonParams -LanguageCode $langCode
                 $null = Update-ListingImage @commonParams -LanguageCode $langCode -RemoveOnly
                 $null = Update-ListingVideo @commonParams -LanguageCode $langCode -RemoveOnly
+                $null = Remove-Listing @commonParams -LanguageCode $langCode
             }
         }
 
--- a/StoreBroker/StoreIngestionListingVideoApi.ps1
+++ b/StoreBroker/StoreIngestionListingVideoApi.ps1
@@ -482,8 +482,6 @@ function Update-ListingVideo
     {
         $stopwatch = [System.Diagnostics.Stopwatch]::StartNew()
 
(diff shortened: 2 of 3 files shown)


StoreBroker/StoreIngestionListingVideoApi.ps1
```diff
@@ -504,6 +502,8 @@ function Update-ListingVideo
 
         if (-not $RemoveOnly)
         {
+            $MediaRootPath = Resolve-UnverifiedPath -Path $MediaRootPath
```
HowardWolosky, line 505: Looks like you weren't sync'd to head.

StoreBroker/StoreIngestionListingApi.ps1
```diff
@@ -836,9 +836,9 @@ function Update-Listing
             Write-Log -Message 'Now removing listings for languages that were cloned by the submission but don''t have current user data.' -Level Verbose
             foreach ($langCode in $listingsToDelete)
             {
-                $null = Remove-Listing @commonParams -LanguageCode $langCode
                 $null = Update-ListingImage @commonParams -LanguageCode $langCode -RemoveOnly
                 $null = Update-ListingVideo @commonParams -LanguageCode $langCode -RemoveOnly
+                $null = Remove-Listing @commonParams -LanguageCode $langCode
```
HowardWolosky, line 841: Strange...I would have thought we would have caught this error years ago.  Good catch

HowardWolosky (approved):

StoreBroker/StoreIngestionListingVideoApi.ps1
HowardWolosky, line 505: Good catch that we needed that change here too.

HowardWolosky:
Don't forget to rev the version number as well.

StoreBroker/StoreIngestionListingApi.ps1
joseartrivera: I am guessing removing a previously included language was a rare occurrence

joseartrivera:
> Don't forget to rev the version number as well.

Good catch!

PR #222, opened 2021-10-19
v2 New-SubmissionPackage: Fix PowerShell 6+ compatibility

DHowett:
PowerShell 6+'s JSON deserializer deserializes numbers as `int64`. By checking for both, we can grow PS6+ compatibility and keep PS<=5.1 compatibility.


DHowett:
please

PR #223, opened 2022-01-19
Fix issues from CodeQL

joseartrivera:
Fixes the following issues:
- Debug flag should not be set to true in Web.config
- Missing x-frame-options from Web.config
- WebRequests are created without setting ServicePointManager.CheckCertificateRevocationList
- Hard-coded use of 'Tls12' security protocol


RESTProxy/Models/ProxyManager.cs
```diff
@@ -116,9 +116,7 @@ public static void Configure(List<Endpoint> endpoints, string defaultTenantId)
             if (!string.IsNullOrWhiteSpace(defaultTenantId))
             {
                 ProxyManager.defaultTenantId = defaultTenantId.ToLowerInvariant();
-            }
-
-            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12;
```
HowardWolosky, line 121: How does removing this mesh with the need when you added it with #191?

joseartrivera: I _think_ it should work by default now that we are targeting .NET Framework 4.8, but I will need to test once I am able to deploy the service to the Cloud Services (Extended Support) instance, which we are currently blocked from doing so until we resolve these CodeQL issues.

HowardWolosky: Seems odd that it's complaining about TLS 1.2.  The documentation (at least for GO and Python) indicate that it considers TLS 1.2 to be fine:

* https://codeql.github.com/codeql-query-help/go/go-insecure-tls/
* https://codeql.github.com/codeql-query-help/python/py-insecure-default-protocol/

joseartrivera: It is not complaining about TLS 1.2, it is complaining about hard coding it because it prevents it from auto-upgrading when a new version releases.

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca5398

PR #229, opened 2022-06-15
Fix telemetry failure for long-running sessions

HowardWolosky:
The telemetry code was creating a singleton base telemetry event the first time that telemetry was being sent.  That base event included the date.  If a PowerShell session lasted multiple days, the telemetry reporting would eventually start to fail because the date being reported for the event was too old.

We now just create a new telemetry event every time one is requested (this also ensures that it honors changes to the `ApplicationInsightsKey`), and now we only cache a GUID that we use to represent the "session" ID which we consider to be the duration of the current PowerShell session.

### DIFF
--- a/StoreBroker.psd1
+++ b/StoreBroker.psd1
@@ -7,7 +7,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '2.1.16'
+    ModuleVersion = '2.1.17'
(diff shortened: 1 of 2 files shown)


StoreBroker.psd1
```diff
@@ -7,7 +7,7 @@
     CompanyName = 'Microsoft Corporation'
     Copyright = 'Copyright (C) Microsoft Corporation.  All rights reserved.'
 
-    ModuleVersion = '2.1.16'
+    ModuleVersion = '2.1.17'
```
joseartrivera, line 10: This is out of date now, otherwise looks good

joseartrivera (approved):

PR #238, opened 2023-10-05
Import global settings from environment

stevenvergenz:
When they're unset, take the ingestion config options from environment variables if available. If not, fill in hard-coded defaults like normal.


StoreBroker/StoreIngestionApi.psm1
```diff
@@ -128,32 +128,76 @@ function Initialize-StoreIngestionApiGlobalVariables
     # SilentlyContinue would cause it to go into the global $Error array, Ignore prevents that as well.
     if (!(Get-Variable -Name SBDefaultProxyEndpoint -Scope Global -ValueOnly -ErrorAction Ignore))
     {
-        $global:SBDefaultProxyEndpoint = $null
+        if ($env:SBDefaultProxyEndpoint)
```
joseartrivera, line 131: Any reason not to uses Get-Variable like we do above?

PR #254, opened 2024-06-06
Fix logging fileSasUri in log file

joseartrivera:
Fixes logging fileSasUri in StoreBroker log file.

Created helper function Write-Body to consolidate filtering logic.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -652,18 +652,18 @@ function Write-Log
 
 $script:alwaysRedactParametersForLogging = @(
     'AccessToken', # Would be a security issue
-    'SasUri'
+    'SasUri' # Could contain a live access token
 )
 
 $script:alwaysRedactHashPropertiesForLogging = @(
-    'fileSasUri'
+    'fileSasUri' # Could contain a live access token
 )
 
 $script:alwaysExcludeParametersForLogging = @(
     'NoStatus'
 )
 
-function Write-Body
+function Write-InputObject
 {
     <#
     .SYNOPSIS
@@ -676,19 +676,21 @@ function Write-Body
         Object to write the body for.
 
     .EXAMPLE
-        Write-Body InputObject $MyObject
+        Write-InputObject InputObject $MyObject
     
     .NOTES
         This method only supports hashtable and PSCustomObject types.
 #>
     [CmdletBinding(SupportsShouldProcess)]
     param(
-        [Object] $InputObject
+        [Object] $InputObject,
+
+        [string] $Description = "Body"
     )
 
     if ($null -eq $InputObject)
     {
-        Write-Log -Message "Body: $null" -Level Verbose
+        Write-Log -Message "$($Description): $null" -Level Verbose
         return
     }
 
@@ -698,7 +700,7 @@ function Write-Body
 
         foreach ($key in @($InputObject.keys))
         {
-            if ($key.ToLower() -in $script:alwaysRedactHashPropertiesForLogging)
+            if ($key -in $script:alwaysRedactHashPropertiesForLogging)
             {
                 $InputObject[$key] = "<redacted>"
             }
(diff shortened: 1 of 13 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -651,13 +651,81 @@ function Write-Log
...
+        return
+    }
+
+    if ($InputObject -is [hashtable])
+    {
+        $InputObject = $InputObject.Clone() # Get a new instance, not a reference
+
+        foreach ($key in @($InputObject.keys))
+        {
+            if ($key.ToLower() -in $script:alwaysRedactHashPropertiesForLogging)
```
HowardWolosky, line 701: PowerShell string comparisons are case-insensitive by default.  You don't _need_ to do that `ToLower()`, but you _can_ if you want.

StoreBroker/Helpers.ps1
```diff
@@ -651,13 +651,81 @@ function Write-Log
 }
 
 $script:alwaysRedactParametersForLogging = @(
-    'AccessToken' # Would be a security issue
+    'AccessToken', # Would be a security issue
+    'SasUri'
```
HowardWolosky, line 655:
```suggestion
    'SasUri' # Could contain a live access token
```

StoreBroker/Helpers.ps1
```diff
@@ -651,13 +651,81 @@ function Write-Log
 }
 
 $script:alwaysRedactParametersForLogging = @(
-    'AccessToken' # Would be a security issue
+    'AccessToken', # Would be a security issue
+    'SasUri'
+)
+
+$script:alwaysRedactHashPropertiesForLogging = @(
+    'fileSasUri'
```
HowardWolosky, line 659:
```suggestion
    'fileSasUri' # Could contain a live access token
```

StoreBroker/Helpers.ps1
```diff
@@ -651,13 +651,81 @@ function Write-Log
...
+
+    .EXAMPLE
+        Write-Body InputObject $MyObject
+    
+    .NOTES
+        This method only supports hashtable and PSCustomObject types.
+#>
+    [CmdletBinding(SupportsShouldProcess)]
+    param(
+        [Object] $InputObject
```
HowardWolosky, line 686: Totally optional depending on how much more time you want to put into this.  To make this more flexible for the future, I would add an additional optional parameter like this:

```powershell
    [string] Description = `Body`,
   [
```
and then do

```powershell
    $message = "$($Description):$objectAsString"
    Write-Log -Message $message -Level Verbose

```

I'd then change the name from `Write-Body` to `Write-InputObject` (open to better naming suggestions, given that `Write-Object` is a real thing).

HowardWolosky (approved):
Overall, I think your approach is totally solid.  Including some suggestions for you to consider.

HowardWolosky:
I didn't review the whole codebase to see if there were any other places you were missing.
You may also want to see if a similar, but different, change needs to be made for v1.

HowardWolosky (changes requested):
I missed the fact that you need to update the version number in the psd1, and should probably update the CHANGELOG as well.

StoreBroker/Helpers.ps1
```diff
@@ -651,13 +651,83 @@ function Write-Log
...
+function Write-InputObject
+{
+    <#
+    .SYNOPSIS
+        Writes a log entry for the passed in object.
+
+    .DESCRIPTION
+        Writes a log entry for the passed in object.
+
+    .PARAMETER InputObject
```
HowardWolosky, line 675: Can you add a `.PARAMETER` for `Description` as well?

HowardWolosky (approved):
Aside from the missing comment, I think this looks great.  Thanks for making the change! (I still don't _love_ the name `Write-InputObject`, but I'm the one that suggested it and I can't think of anything better.  Naming is hard.)

PR #257, opened 2024-06-10
Refactor logic to exclude logging of fileSasUri in parameter objects

joseartrivera:
Need to not log filesas objects in Write-InvocationLog when they are the properties of PSCustomObjects or Hashtables that are part of the passed in parameters.

Factor out the existing logic in Write-InputObject into a helper function that only does the copy & replace of the object.

Use that in Write-InvocationLog whenever a parameter is a PSCustomObject or Hashtable.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -663,6 +663,66 @@ $script:alwaysExcludeParametersForLogging = @(
     'NoStatus'
 )
 
+function Copy-InputObjectForLogging
+{
+    <#
+    .SYNOPSIS
+        Creates a copy of the InputObject and redacts some of its contents for logging.
+
+    .DESCRIPTION
+        Creates a copy of the InputObject and redacts some of its contents for logging.
+
+    .PARAMETER InputObject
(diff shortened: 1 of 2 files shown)


StoreBroker/Helpers.ps1
```diff
@@ -663,71 +663,108 @@ $script:alwaysExcludeParametersForLogging = @(
     'NoStatus'
 )
 
-function Write-InputObject
+function Copy-InputObjectForLogging
```
adstep, line 666: Is this method expected to handle nested properties? We might need to introduce some recursion here to ensure everything is caught.

joseartrivera: Just covering the applicable use-cases in this PR. We can update it to be all-encompassing if need be at some point, but not necessary to fix the current issue.

adstep (approved):

HowardWolosky (commented):
This merged in before I had a chance to complete providing my feedback.  I think that there are likely some things in this you may want to consider addressing in a follow-up PR.

PR #264, opened 2024-06-13
Reintroduce deep copy fix

joseartrivera:
In #263 we used DeepCopy-Object to create a copy of hashtables and pscustomobjects. This caused enums to be treated as PSCustomObjects and would later throw exceptions when Convert-EnumToString would call the copy() function on it.

My initial thought to fix was to call Convert-EnumToString before doing the deep-copy. However I noticed Convert-EnumToString is accomplishing the same thing we want to do and instead of using DeepCopy it does a shallow copy with Copy() and Clone() and then recursively makes more shallow copies.

We can modify our original solution to do the same thing since we are also calling child objects recursively if we plan on modifying them.

### DIFF
--- a/StoreBroker/Helpers.ps1
+++ b/StoreBroker/Helpers.ps1
@@ -691,39 +691,31 @@ function Copy-InputObjectForLogging
 
     if ($InputObject -is [hashtable])
     {
-        $replacedInputObject = DeepCopy-Object -InputObject $InputObject # Get a new instance, not a reference
+        $ReplacedInputObject = $InputObject.Clone() # Get a new instance, not a reference
 
-        foreach ($key in @($replacedInputObject.Keys))
+        foreach ($key in @($ReplacedInputObject.keys))
         {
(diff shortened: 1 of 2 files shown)


adstep (approved):

StoreBroker/Helpers.ps1
```diff
@@ -754,18 +762,18 @@ function Write-InputObject
 
     if ($null -eq $InputObject)
     {
-        Write-Log -Message "$($Description): $null" -Level Verbose
+        Write-Log -Message "$($Description): `$null" -Level Verbose
```
HerrickSpencer, line 765: Nit: is there a reason to even include the dollar sign here?  it is just a log line.

HerrickSpencer (approved):

StoreBroker/Helpers.ps1
joseartrivera: It is more clear to the reader that the return is null since in powershell $null = null

PR #265, opened 2024-06-24
refactor: Improve path handling and code readability

kochie:
This PR fixes an issue where on platforms like Unix and MacOS the submission script will fail due to incorrect path separators. The code now uses the system path separators to determine which language to use.


kochie:
@microsoft-github-policy-service agree

PR #275, opened 2025-03-27
Return new HttpResponseMessage instead of one from the service call in EndPoint.cs

joseartrivera:
Prevents any information exposure from the APIs being forwarded.

Note: This change will lose client-side access to correlationId for RestProxy, consider a better method of not exposing additional information in the future.


HowardWolosky:
Keep in mind that this change means that you'll lose access to the `correlationId` which would be required for server-side investigation.

adstep (approved):

